PR #9, opened 2017-11-10
Add filename param to main interface

### LINKED ISSUE #8: Inability to set source filename
For tooling, it is important to have the file names in order to print warnings/errors.

Currently, `asttokens.ASTTokens` does not allow to pass in the file name it's parsing.

It should be possible to do that here: https://github.com/gristlabs/asttokens/blob/master/asttokens/asttokens.py#L49

LiraNuna:
This PR adds the parameter `filename` to the main ASTToken interface.

I took some liberty making some decisions that may have undesired consequences:

~1) I made `filename` a param after `source_text` to better mimic `ast.parse`'s interface - this will cause issues if someone used this library without kwargs. If this was python3 I would've added `*` between `filename` and `parse` to enforce this going forward~
2) I duplicated `filename` on the instance since the tree can be passed independently (see line 50)
3) I added a getter because why not

Fixes #8

### DIFF
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -42,11 +42,12 @@ class ASTTokens(object):
   If only ``source_text`` is given, you may use ``.mark_tokens(tree)`` to mark the nodes of an AST
   tree created separately.
   """
-  def __init__(self, source_text, parse=False, tree=None):
+  def __init__(self, source_text, parse=False, tree=None, filename='<unknown>'):
     if isinstance(source_text, six.binary_type):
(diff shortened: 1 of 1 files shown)


dsagal:
Thank you! Let's just move `filename` to the end. Everyone *should* use named arguments, and for those people it doesn't matter (question of consistency with `ast.parse` is moot too); but if anyone uses positional arguments today, we don't want to break their code unnecessarily.

LiraNuna:
@dsagal changes made.

dsagal:
Thank you!

PR #13, opened 2018-02-15
Handle JoinedStr nodes (f-strings in Python 3)

FuegoFro:
This diff adds special handling for `JoinedStr` nodes (which are produced by f-strings added in Python 3.6). This handling mirrors what is already being done for normal strings. Before, the `source` attribute of a `JoinedStr` node would only have the first token if the string used compile-time concatenation; now it will have all the tokens in that case.

I added a test that mirrors the existing `Str` test, saw that it failed, implemented the fix, and saw that it now succeeds. I ran the test on Python 2.7 and 3.6 and saw that it passed on both. I temporarily took out the guard at the beginning of the test and ran in on Python 2 and saw that it failed with a syntax error.

### DIFF
--- a/asttokens/mark_tokens.py
+++ b/asttokens/mark_tokens.py
@@ -230,6 +230,12 @@ class MarkTokens(object):
     return (first_token, last_token)
 
   def visit_str(self, node, first_token, last_token):
+    return self.handle_str(first_token, last_token)
+
+  def visit_joinedstr(self, node, first_token, last_token):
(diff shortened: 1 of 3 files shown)


FuegoFro:
Oh nice, much cleaner :D

dsagal:
Thank you @FuegoFro ! I'll publish an updated version now.

FuegoFro:
Thank *you*!

PR #15, opened 2018-02-16
Adjust With nodes to include the "with" text.

dsagal:
Apparently in Python 2 the `col_offset` of `With` nodes starts after the "with " text, which is incorrect. It works correctly on Python 3. This PR fixes it for Python 2.

Builds on PR #14 by @FuegoFro, with a different implementation and more testing.

### DIFF
--- a/asttokens/mark_tokens.py
+++ b/asttokens/mark_tokens.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import six
-import sys
 import numbers
 import token
 from . import util
--- a/asttokens/mark_tokens.py
+++ b/asttokens/mark_tokens.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import six
+import sys
 import numbers
 import token
 from . import util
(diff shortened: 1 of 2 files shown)


asttokens/mark_tokens.py
```diff
@@ -279,3 +280,9 @@ def visit_assignname(self, node, first_token, last_token):
       colon = self._code.find_token(last_token, token.OP, ':')
       first_token = last_token = self._code.prev_token(colon)
     return (first_token, last_token)
+
+  if six.PY2:
```
FuegoFro: Oh, that's certainly the better way to do the check haha. Plus, the `import sys` above is no longer needed :D

tests/test_mark_tokens.py
```diff
@@ -448,3 +448,36 @@ def g(x):
...
+      "Module:with foo: pass",
+      "With:with foo: pass",
+    })
+
+    source = textwrap.dedent(
+      '''
+      def f(x):
+        with A() as a:
+          log(a)
+          with B() as b, C() as c: log(b, c)
```
FuegoFro: Oh nice, I should have tested the multi-with case. Super interesting that it desugars it to be nested `with` nodes in Python 2, but a single `with` node with multiple items in Python 3 :)

Issue #21, opened 2018-10-06
ASTTokens won't parse strings with coding declarations in Python 2


ssbr:
Python 2:

```python
>>> asttokens.ASTTokens("# coding: ascii\n1", parse=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/google/home/devinj/.local/lib/python2.7/site-packages/asttokens/asttokens.py", line 50, in __init__
    self._tree = ast.parse(source_text, filename) if parse else tree
  File "/usr/lib/python2.7/ast.py", line 37, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 0
SyntaxError: encoding declaration in Unicode string
```

In Python 3 it works just fine.

I expect failures if the string is not UTF-8 (e.g. if it had coding:latin1 and some invalid utf-8 in there), but I had somewhat hoped that it would also just work to pass in the raw bytes.

I think the easiest workaround for end users is to parse it manually by hand:

```python
>>> s = "# coding: ascii\n1"
>>> asttokens.ASTTokens(s, tree=ast.parse(s))
<asttokens.asttokens.ASTTokens object at 0x7f35d3b76550>
```

asttokens could also do that internally itself -- parse the tree before decoding, instead of after.

ssbr:
I'm not very good at git/github (use a different VCS at work), so I'm not sure how everything works, but I made a PR with my suggested fix: https://github.com/gristlabs/asttokens/pull/22

alexmojaki:
This is fixed by #22 now, right?

ssbr:
Yes, I believe so. (Sorry about that; not used to github issues etc., we don't use github as much at work.)

PR #22, opened 2018-10-06
Per issue #21, delay decoding the source code until _after_ parsing.

ssbr:
Because Python 2 forbids coding declarations inside of unicode strings, passing unicode/text strings to ast.parse makes all files with coding declarations unparseable with asttokens in Python 2.

### DIFF
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -43,11 +43,14 @@ class ASTTokens(object):
   tree created separately.
   """
   def __init__(self, source_text, parse=False, tree=None, filename='<unknown>'):
-    if isinstance(source_text, six.binary_type):
-      source_text = source_text.decode('utf8')
-
     self._filename = filename
     self._tree = ast.parse(source_text, filename) if parse else tree
+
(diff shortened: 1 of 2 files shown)


dsagal:
Thank you, looks great. I have two requests before landing.

One to rebase to tip of master (I fixed tox requirements so that tests should pass after you rebase). 

The other is about the test. I don't see the purpose of `str()` in the call (it seems to negate the comment about unicode in Python 2), so it seems clearer to remove it (or add a comment why it belongs there).

And it would be nice to assert that the sample gets parsed as expected -- you can do that by assigning the result to `atok` and then doing `self.assertEqual([str(t) for t in atok.tokens], [...])` (as in some other test cases).

ssbr:
Sorry about the delay, busy week. I don't think it made sense to have the str() call so I removed it. (The intent had been to test bytes on Python 2, and unicode on Python 3, but there's no reason not to test bytes on both.) Pulled/rebased as requested.

ssbr:
I'm pretty much a github novice (we use a different VCS at work). Is there anything else I need to do on my end in order to get this merged?

PR #29, opened 2019-04-18
Fix some Python 3.8 issues

vstinner:
* tox.ini: add py38
* tests.tools: add Constant() helper

Partially addresses #28

### DIFF
--- a/tests/tools.py
+++ b/tests/tools.py
@@ -132,3 +132,14 @@ class MarkChecker(object):
       tested_nodes += 1
 
     return tested_nodes
+
+
+def Constant(name):
+  """Helper to write tests working on all Python versions.
+
+  Return 'Constant' on Python 3.8 and newer, or name on older Python.
+  """
+  if sys.version_info >= (3, 8):
+    return 'Constant'
+  else:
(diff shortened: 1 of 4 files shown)


tests/tools.py
```diff
@@ -132,3 +132,14 @@ def verify_all_nodes(self, test_case):
       tested_nodes += 1
 
     return tested_nodes
+
+
+def Constant(name):
```
dsagal: What do you think of changing `MarkChecker.view_node` to replace `node.__class__.__name__` with `get_node_name(node)` which is something like this:
```
def get_node_name(node):
   name = node.__class__.__name__
   return "Constant" if name in ("Num", "Str", "NameConstant") else name
```
Then we can change code throughout to expect things like `Constant:17` or `Constant:'hello'`, which seems a bit simpler.

vstinner:
Sorry, I don't have the bandwidth to update my PR. If someone wants to take it over, please go ahead :-)

keszybz:
See #30.

keszybz:
This can be closed now.

PR #36, opened 2019-10-04
Python 3.8 fixes

alexmojaki:
Fixing test failures noted in #28. Still working through the remaining failures so don't rush to merge, just letting you know about progress.

### DIFF
--- a/asttokens/mark_tokens.py
+++ b/asttokens/mark_tokens.py
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import six
 import numbers
+import sys
 import token
-from . import util
 
+import six
+
+from . import util
 
 # Mapping of matching braces. To find a token here, look up token[:2].
 _matching_pairs_left = {
@@ -171,8 +173,9 @@ class MarkTokens(object):
     util.expect_token(before, token.OP, open_brace)
     return (before, last_token)
 
-  def visit_listcomp(self, node, first_token, last_token):
-    return self.handle_comp('[', node, first_token, last_token)
+  if sys.version_info[:2] < (3, 8):
+    def visit_listcomp(self, node, first_token, last_token):
+      return self.handle_comp('[', node, first_token, last_token)
 
   if six.PY2:
     # We shouldn't do this on PY3 because its SetComp/DictComp already have a correct start.
--- a/asttokens/mark_tokens.py
+++ b/asttokens/mark_tokens.py
@@ -202,15 +202,17 @@ class MarkTokens(object):
   visit_assignattr = handle_attr
   visit_delattr = handle_attr
 
-  def handle_doc(self, node, first_token, last_token):
+  def visit_functiondef(self, node, first_token, last_token):
     # With astroid, nodes that start with a doc-string can have an empty body, in which case we
     # need to adjust the last token to include the doc string.
     if not node.body and getattr(node, 'doc', None):
       last_token = self._code.find_token(last_token, token.STRING)
+    prev = self._code.prev_token(first_token)
+    if prev[:2] == (token.OP, '@'):
+      first_token = prev
     return (first_token, last_token)
 
-  visit_classdef = handle_doc
-  visit_funcdef = handle_doc
+  visit_classdef = visit_functiondef
 
   def visit_call(self, node, first_token, last_token):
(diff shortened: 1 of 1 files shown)


alexmojaki:
OK that fixes all the tests except test_fixture8, test_mark_tokens_multiline, and test_slices, all of which are failing because tuples now include parentheses, which needs to be discussed.

asttokens/mark_tokens.py
```diff
@@ -171,8 +173,9 @@ def handle_comp(self, open_brace, node, first_token, last_token):
     util.expect_token(before, token.OP, open_brace)
     return (before, last_token)
 
-  def visit_listcomp(self, node, first_token, last_token):
-    return self.handle_comp('[', node, first_token, last_token)
+  if sys.version_info[:2] < (3, 8):
```
dsagal: Do we really want to exclude this? If so, could you add a comment to explain? Also, isn't it enough to compare `sys.version_info < (3, 8)` (without taking a slice)?

asttokens/mark_tokens.py
```diff
@@ -199,15 +202,17 @@ def handle_attr(self, node, first_token, last_token):
...
   visit_delattr = handle_attr
 
-  def handle_doc(self, node, first_token, last_token):
+  def visit_functiondef(self, node, first_token, last_token):
     # With astroid, nodes that start with a doc-string can have an empty body, in which case we
     # need to adjust the last token to include the doc string.
     if not node.body and getattr(node, 'doc', None):
       last_token = self._code.find_token(last_token, token.STRING)
+    prev = self._code.prev_token(first_token)
+    if prev[:2] == (token.OP, '@'):
```
dsagal: Cleaner (and more consistent with other code) to use `if util.match_token(prev, token.OP, '@')`

asttokens/mark_tokens.py
```diff
@@ -199,15 +202,17 @@ def handle_attr(self, node, first_token, last_token):
   visit_assignattr = handle_attr
   visit_delattr = handle_attr
 
-  def handle_doc(self, node, first_token, last_token):
+  def visit_functiondef(self, node, first_token, last_token):
     # With astroid, nodes that start with a doc-string can have an empty body, in which case we
     # need to adjust the last token to include the doc string.
     if not node.body and getattr(node, 'doc', None):
       last_token = self._code.find_token(last_token, token.STRING)
+    prev = self._code.prev_token(first_token)
```
dsagal: A bit concerned that if `first_token` is the first-ever token, then `prev_token` throws. To guard against it, I suggest wrapping into `if first_token.index > 0:`

asttokens/mark_tokens.py
```diff
@@ -199,15 +202,17 @@ def handle_attr(self, node, first_token, last_token):
...
+    if prev[:2] == (token.OP, '@'):
+      first_token = prev
     return (first_token, last_token)
 
-  visit_classdef = handle_doc
-  visit_funcdef = handle_doc
+  visit_classdef = visit_functiondef
```
dsagal: It reads a bit misleading to use `visit_functiondef` for `classdef`. Maybe call the helper method `handle_def`, to clarify that it's not specific?

asttokens/mark_tokens.py
```diff
@@ -261,6 +266,8 @@ def visit_const(self, node, first_token, last_token):
       return self.visit_str(node, first_token, last_token)
     return (first_token, last_token)
 
+  visit_constant = visit_const
```
dsagal: Could you add a comment about this too?

asttokens/mark_tokens.py
alexmojaki: `col_offset` was changed for list comprehensions and tuples in 3.8: https://bugs.python.org/issue31241

Code to try it out:

```python
import ast

source = "[x for x in y]"
tree = ast.parse(source)
comp = tree.body[0].value
print(comp, comp.col_offset)
```

asttokens/mark_tokens.py
alexmojaki: It doesn't throw, it returns something like `tokens[-1]`. Which is probably a problem that we should fix, just not here. But I can add that check anyway, particularly in case we make it throw in the future.

alexmojaki:
I'd like to leave the issue of tuples and parentheses out of this PR so that less code is hanging around in parentheses.

PR #38, opened 2019-10-04
Support pypy

alexmojaki:


asttokens/mark_tokens.py
```diff
@@ -251,6 +252,15 @@ def visit_num(self, node, first_token, last_token):
     # A constant like '-1' gets turned into two tokens; this will skip the '-'.
     while util.match_token(last_token, token.OP):
       last_token = self._code.next_token(last_token)
+
+    # This makes sure that the - is included
+    if first_token.type == token.NUMBER:
+      if isinstance(node, ast.AST):
```
dsagal: I have a suggestion that I think improves on isinstance here. Rename this method to be a helper, like `handle_num(node, value, first_token, last_token)`, with `value` as an extra argument. Then the new condition becomes simply:

```
if value < 0 and first_token.type == token.NUMBER:
  first_token = self._code.prev_token(first_token)
```

Then from `visit_num()`, call `handle_num(node, node.n, first_token, last_token)`, and from `visit_const()` call `handle_num(node, node.value, first_token, last_token)`.

alexmojaki: That won't work in 3.8 when the number has type `Constant` which calls `visit_const` after #36. Plus I think it's good that it's explicit, rather than hiding logic in method names.

dsagal: We are just talking about whether to use `node.value` or `node.n`, right?

What I meant is that as long as `visit_const` is called, it already pays attention to `node.value`, so that's the guaranteed correct value to use. On the other hand, in `visit_num`, `node.n` is the correct value to use (always, as far as I can tell). I feel that's more explicit (it's similar to checking whether the node's type is `Num`, which seems more explicit than `isinstance(node, ast.AST)`). Did I misunderstand?

alexmojaki: You're right, I was thinking that with `ast` I always have to use `.n`, but `ast.Constant` has `.value` too.

PR #39, opened 2019-10-07
Find parentheses in tuples, fully supporting 3.8

alexmojaki:
Will add some more tests tomorrow.

### DIFF
--- a/asttokens/mark_tokens.py
+++ b/asttokens/mark_tokens.py
@@ -219,13 +219,39 @@ class MarkTokens(object):
     return (first_token, self._code.find_token(last_token, token.OP, ']'))
 
   def visit_tuple(self, node, first_token, last_token):
-    # A tuple doesn't include parens; if there is a trailing comma, make it part of the tuple.
-    try:
-      maybe_comma = self._code.next_token(last_token)
-      if util.match_token(maybe_comma, token.OP, ','):
-        last_token = maybe_comma
-    except IndexError:
-      pass
+    # Tuples may or may not contain parentheses
+    # If they do, we want to include them
+    # We may or may not already have them, we need to find out
+    # We have them if the tokens start and end with matching ()
+    # meaning if we remove them we are still left with matching parens
+    contains_parens = False
+    if util.is_parentheses(first_token, last_token):
+      ops = [
+              tok
+              for tok in self._code.token_range(first_token, last_token)
+              if tok.type == token.OP
+            ][1:-1]
+      if util.matched_paretheses(ops):
+        contains_parens = True
+
+    # If there is a trailing comma, make it part of the tuple.
+    if not contains_parens:
+      try:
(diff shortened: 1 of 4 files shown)


asttokens/mark_tokens.py
```diff
@@ -219,13 +219,39 @@ def visit_subscript(self, node, first_token, last_token):
...
-        last_token = maybe_comma
-    except IndexError:
-      pass
+    # Tuples may or may not contain parentheses
+    # If they do, we want to include them
+    # We may or may not already have them, we need to find out
+    # We have them if the tokens start and end with matching ()
+    # meaning if we remove them we are still left with matching parens
+    contains_parens = False
+    if util.is_parentheses(first_token, last_token):
```
dsagal: I don't think it's OK to decide on the spot this way. A tuple like `((1), (2))` will present itself as `(1), (2)` pre-3.8, and as `((1), (2))` post-3.8, but in both cases we start and end with parentheses. I think  we need to have a version check, and expand to include the parens if pre-3.8. (Or alternatively, skip the paren in >=3.8, if it might be better to keep parens out, to maintain past behavior.)

alexmojaki: That's handled in the following lines as explained in the preceding comment

dsagal: Sorry, I think I understand your point a bit better now, I need to think about it. It seems overly complicated.

dsagal:
I think an important case to consider is this: `((foo, bar), )`. I suspect that it's not possible to treat consistently without taking the python version into consideration.

alexmojaki:
That's a great point. It gets worse if you consider `(foo, bar),`. Assuming you have the inner tuple correct it doesn't matter what the version is. But it looks like there are already parentheses so it won't try to include the trailing comma. I think I need to add another check 'if there's only one element and it has the same first and last token, parens have not been included'. That should also handle your example. If there's more than one, e.g. `(foo,), (bar,),`, the matching parens trick will come into play. But I agree this is quite complicated and I would be happier if there was a simpler method that worked.

dsagal:
How do you feel about logic like this?
1. if python < 3.8 and if prev-token is "(", then set first-token to
prev-token (skip this step for >= 3.8)
2. if first-token is "(", then expand to matching parentheses; else use
existing logic for trailing comma

alexmojaki:
I've been testing my own algorithm and learned that an additional complication to consider is parentheses unrelated to tuples, like `((a),)`. That will mess with your algorithm. Apart from that I'll need more time to think about it.

dsagal:
Yikes, in code like `(a,),`, Python 3.8 considers both nested tuples to start with the same opening paren. One could consider the first-child of a tuple, but both `(a),` and `(a,)` have `a` as the first-child, so that idea doesn't help much :(

dsagal:
Could you take a look at #40 please? I think I managed to handle the bothersome examples, but it's hard to be sure.

PR #40, opened 2019-10-09
Have tuples include parentheses when present, and support Python 3.8.

dsagal:


alexmojaki:
Haven't had a chance to read through it, but well done anyway! Here are the tests I put together for myself:

```python
  def test_tuples(self):
    def check(source, *extra_tuples):
      m = self.create_mark_checker(source)
      self.assertEqual(
        {
          m.atok.get_text(node)
          for node in m.all_nodes
          if node.__class__.__name__ == 'Tuple'
        },
        {source}.union(extra_tuples)
      )
      m.verify_all_nodes(self)

    check('()')
    check('(),', '()')
    check('a,')
    check('(a,)')
    check('((a),)')
    check('((),)', '()')
    check('((),(a,))', '()', '(a,)')
    check('((),(a,),)', '()', '(a,)')
    check('((),(a,),),', '()', '(a,)', '((),(a,),)')
    check('((foo, bar),)', '(foo, bar)')
    check('(foo, bar),', '(foo, bar)')
```

dsagal:
I added the examples from your test case, and a bit more. Seems robust so far.

asttokens/mark_tokens.py
```diff
@@ -228,6 +228,41 @@ def visit_tuple(self, node, first_token, last_token):
       pass
     return (first_token, last_token)
 
+  if sys.version_info >= (3, 8):
+    # In Python3.8 parsed tuples include parentheses when present.
+    def handle_tuple_nonempty(self, node, first_token, last_token):
+      # It's a bare tuple if the first token belongs to the first child. The first child may
+      # include extraneous parentheses (which don't create new nodes), so account for those too.
+      child = next(self._iter_children(node))
```
alexmojaki: Why not `node.elts[0]`? This threw me off for a moment.

alexmojaki:
Great, I think your algorithm makes sense. I'm just wondering if it's possible to have any tokens between the tuple first token and the child first token other than `(` and non-coding tokens, but I can't think of anything. I'd go ahead with this.

alexmojaki:
Let's merge this. 3.8 is scheduled for release today.

dsagal:
Published release 2.0.0, bumping major version because this commit includes a slight change to parsing results.

PR #41, opened 2019-10-15
Test sys.modules

alexmojaki:
Looks like there's still lots of work to do.

Will push a fix for the decorator bug just now, after the travis build completes so you can see the failure log.

### DIFF
--- a/tests/test_mark_tokens.py
+++ b/tests/test_mark_tokens.py
@@ -1,11 +1,16 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals, print_function
-import astroid
-import six
+
+import inspect
+import io
+import os
 import sys
-import token
 import textwrap
 import unittest
+
+import astroid
+import six
+
 from . import tools
 
 
@@ -473,6 +478,10 @@ bar = ('x y z'   # comment2
       @deco2(a=1)
       def g(x):
         pass
+
+      @deco3()
+      def g(x):
+        pass
     """)
     m = self.create_mark_checker(source)
     m.verify_all_nodes(self)
@@ -550,3 +559,21 @@ bar = ('x y z'   # comment2
     source = 'f((x)[:, 0])'
     m = self.create_mark_checker(source)
     m.verify_all_nodes(self)
+
+  def test_sys_modules(self):
+    for module in list(sys.modules.values()):
+      try:
+        filename = inspect.getsourcefile(module)
+      except TypeError:
+        continue
+
+      if not filename:
+        continue
+
+      filename = os.path.abspath(filename)
+      print(filename)
+      with io.open(filename) as f:
+        source = f.read()
+      source = source.replace('elif', 'if')
+      m = self.create_mark_checker(source)
+      m.verify_all_nodes(self)


tests/test_mark_tokens.py
```diff
@@ -550,3 +561,21 @@ def test_complex_slice_and_parens(self):
...
+        continue
+
+      if not filename:
+        continue
+
+      filename = os.path.abspath(filename)
+      print(filename)
+      with io.open(filename) as f:
+        source = f.read()
+      source = source.replace('elif', 'if')
```
dsagal: what's this about?

tests/test_mark_tokens.py
```diff
@@ -550,3 +561,21 @@ def test_complex_slice_and_parens(self):
...
+      try:
+        filename = inspect.getsourcefile(module)
+      except TypeError:
+        continue
+
+      if not filename:
+        continue
+
+      filename = os.path.abspath(filename)
+      print(filename)
```
dsagal: print calls probably don't belong in a test case

asttokens/mark_tokens.py
```diff
@@ -212,6 +212,12 @@ def visit_call(self, node, first_token, last_token):
     call_start = self._code.find_token(first_child.last_token, token.OP, '(')
     if call_start.index > last_token.index:
       last_token = call_start
+
+    # Handling a python bug with decorators with empty parens, e.g.
+    # @deco()
+    # def ...
+    if util.match_token(first_token, token.OP, '@'):
```
dsagal: Really, the AST considers the start of the call node to be the "@" token? That's disappointing..

alexmojaki:
Here's that log: https://travis-ci.org/gristlabs/asttokens/jobs/598019384

I'm seeing a bunch of bugs from different places.

> AttributeError: 'SyntaxError' object has no attribute 'filename'

Maybe a bug in nose? It doesn't happen to me locally

```
    (@deco3())
     ^
SyntaxError: invalid syntax
```

That's the bug that was just fixed. This should probably be reported to Python. Here's a demo:

```python
import ast

source = """
@a()
def b(): 0

@a(x)
def b(): 0

@a
def b(): 0
"""

body = ast.parse(source).body
for fdef in body:
    call = fdef.decorator_list[0]
    print(call.col_offset)
```

Then there's a problem that only appears on the astroid side:

```
- class IOBase(_io._IOBase):
+ class IOBase(_io._IOBase, metaclass=ABCMeta):
      __doc__ = _io._IOBase.__doc__
```

Maybe a bug in astroid?

tests/test_mark_tokens.py
alexmojaki: Maybe, but it's very useful in development and even when tests pass it's quite nice to see all the files that were tested. This is what I've done in my own library.

tests/test_mark_tokens.py
alexmojaki: Sorry, I should probably add a comment. `elif:` is really just `else: if:` to the AST, so `get_text` can return text starting with `elif` when given an `If` node. I think this is generally harmless and there's probably no good alternative, but it leads to syntax errors in `verify_all_nodes`.

alexmojaki:
OK, I need to get going. Feel free to make changes, you automatically have write access to this branch. I know there are at least two more issues in asttokens that the new test will reveal.

tests/test_mark_tokens.py
dsagal: How about, in `verify_all_nodes`, if the text starts with `/^\s*elif\s`, replace "elif" with "if"? (With your comment to explain.) It would be clearer that this is specific to how nodes are verified, and would apply in all cases, not just in one test case.

alexmojaki:
Made an issue in astroid: https://github.com/PyCQA/astroid/issues/705

alexmojaki:
I've created a PR on top of this PR, check out https://github.com/alexmojaki/asttokens/pull/1

alexmojaki:
I'm running into various unicode-related errors in Python 2, and I don't have the will to fix them. It's probably time to phase out Python 2 support anyway.

asttokens/mark_tokens.py
alexmojaki: Made a bug: https://bugs.python.org/issue38535

dsagal:
I published a new version to pypi. Thank you, @alexmojaki!

PR #42, opened 2019-10-24
Coveralls, pytest, and faster astroid tests

alexmojaki:
This is about various testing infrastructure improvements:

1. pytest instead of nose:

Whenever `ast.parse` fails with a SyntaxError inside a test, nose breaks and instead I get:

> AttributeError: 'SyntaxError' object has no attribute 'filename'

The problem doesn't happen with pytest and I'm getting more informative tracebacks too. nose is old now and [its own docs recommend not using it](https://nose.readthedocs.io/en/latest/). I think this would be a great improvement.

2. Coveralls: Travis now sends a coverage report to coveralls. See an example here: https://coveralls.io/builds/26516443

If you merge this you just need to sign up for coveralls and turn the repo on. Then you can add a badge to the README.

3. Speeding up astroid tests: the old astroid `assert_nodes_equal` builds up large strings just to compare them, and it's quite wasteful. Using the same algorithm for both to compare trees directly is significantly faster. This is important because the coverage tracer adds a lot of overhead, and it seems to be especially bad for pypy, which was exceeding the maximum allowed time of 30 minutes on travis. Even now it just barely makes it and we may need to place a time limit, something like [this](https://github.com/alexmojaki/asttokens/commit/45829e6343c0c00bdb32fd2427f13dd4ed37e788).

If you're happy with all this, I will also update the docs regarding testing.

I would also need to update `requirements[-top].txt`, but really I think it would be better to delete them. Having pinned versions (e.g. `astroid==1.5.3`!) is definitely bad, but even otherwise you can just do `pip install .[test]`.

### DIFF
--- a/tests/test_astroid.py
+++ b/tests/test_astroid.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals, print_function
 import re
 
 import astroid
+
 from . import tools, test_mark_tokens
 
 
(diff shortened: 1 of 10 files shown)


tests/test_astroid.py
```diff
@@ -12,38 +13,27 @@ class TestAstroid(test_mark_tokens.TestMarkTokens):
   is_astroid_test = True
   module = astroid
 
+  nodes_classes = astroid.ALL_NODE_CLASSES
```
dsagal: Why is `astroid.NodeNG` not sufficient? (if it really is not, would you mind adding a comment?)

alexmojaki: `astroid.NodeNG` doesn't resolve, it has to be imported from `astroid.node_classes`, and I just find it less readable anyway (what does NodeNG mean?)

alexmojaki:
OK, I think I've done everything we've talked about, assuming the build passes this should be ready to merge.

tests/test_mark_tokens.py
```diff
@@ -592,7 +593,23 @@ def test_complex_slice_and_parens(self):
...
+      """
+      modules = list(sys.modules.values())
+      if not os.environ.get('ASTTOKENS_SLOW_TESTS'):
+        modules = modules[:20]
+
+      start = time()
+      for module in modules:
+        # Don't let this test (which runs twice) take longer than 13 minutes
+        # to avoid the travis build time limit of 30 minutes
+        if time() - start > 13 * 60:
```
dsagal: I'm approving and merging but I find this problematic. Even if ASTTOKENS_SLOW_TESTS is set, tests may pass but there is no telling if it tested all files or 20 files or 2 files.

PR #43, opened 2019-10-24
Automatically call verify_all_nodes by default

alexmojaki:
This is branched from the branch in #42 so ignore all commits except the last one for now, they should disappear when #42 is merged.


alexmojaki:
Try restarting this job to fix the build: https://travis-ci.org/gristlabs/asttokens/jobs/602465044

dsagal:
Looks good, thank you!

alexmojaki:
Can you merge this too? The only failing check is coverage, that problem should be fixed now.

PR #46, opened 2019-10-26
Increase test coverage

alexmojaki:


asttokens/util.py
```diff
@@ -167,8 +162,6 @@ def visit_tree(node, previsit, postvisit):
 
   For the initial node, ``par_value`` is None. Either ``previsit`` and ``postvisit`` may be None.
```
dsagal: It seems "``previsit`` may be None" comment is no longer accurate.

asttokens/mark_tokens.py
```diff
@@ -99,10 +99,7 @@ def _visit_after_children(self, node, parent_token, token):
     node.last_token = nlast
 
   def _find_last_in_line(self, start_token):
-    try:
-      newline = self._code.find_token(start_token, token.NEWLINE)
-    except IndexError:
```
dsagal: Mmmm, I have a feeling removing this `except` can easily cause an error. Why do you think it's better to remove?

asttokens/mark_tokens.py
```diff
@@ -237,12 +234,9 @@ def visit_subscript(self, node, first_token, last_token):
 
   def handle_bare_tuple(self, node, first_token, last_token):
     # A bare tuple doesn't include parens; if there is a trailing comma, make it part of the tuple.
-    try:
-      maybe_comma = self._code.next_token(last_token)
-      if util.match_token(maybe_comma, token.OP, ','):
-        last_token = maybe_comma
-    except IndexError:
```
dsagal: Same here.

tests/test_asttokens.py
```diff
@@ -137,5 +137,24 @@ def test_coding_declaration(self):
...
     ])
 
+
+def test_filename():
+  filename = "myfile.py"
+  atok = asttokens.ASTTokens("a", parse=True, filename=filename)
+  assert filename == atok.filename
+
+
+def test_doesnt_have_location():
```
dsagal: Could you add a comment what this tests, please? The test name isn't self-explanatory.

asttokens/mark_tokens.py
alexmojaki: I'm removing code that never runs. `find_token` doesn't advance past the end because it has a check `and not token.ISEOF(t.type)`. We should maintain that guarantee and also trust it.

asttokens/mark_tokens.py
alexmojaki: `next_token` can't raise an error here because there's always an endmarker in tokens, which I've just added a check for. If `last_token` is already the endmarker then that's a problem which should be fixed. I think our extensive tests, particularly for tuples, are sufficient.

alexmojaki:
Hey, you approved the changes, but I think you forgot you need to merge. Also please release afterwards. There's an actual bug that's been fixed here (the use of `token.ASYNC`) that's causing some failures [here](https://travis-ci.org/ipython/ipython/jobs/626372798?utm_medium=notification&utm_source=github_status).

dsagal:
You got it.

PR #49, opened 2020-01-05
Fix docstring of util.replace

alexmojaki:


alexmojaki:
Hi @dsagal, this is a bit weird but I didn't know where else to put this. I thought you'd be interested to see my latest and greatest project: https://futurecoder.io/

It uses asttokens quite a lot via, mostly other libraries of mine, including snoop, birdseye, and stack_data. Plus all sorts of other magic and introspection with and without the AST.

dsagal:
That's fantastic! I like it, and I'm giving it to a newbie to try :) I may reach out in the app or via email with some feedback. Congratulations on launching this!

PR #51, opened 2020-02-12
Moved the metadata into setup.cfg

KOLANICH:


dsagal:
You seem to know what you are doing, and this certainly looks cleaner, thank you! Could you include some pointers about this approach? Is this what's recommended? Other packages structured this way? Does it work fine with Python 2?

KOLANICH:
`setup.cfg` is available since 2016, so is `setuptools_scm`, but support of `pyproject.toml` in `setuptools_scm` was implemented just recently. `setuptools` still doesn't take metadata from `pyproject.toml`, but when it will be fixed (I guess not very soon), we'll he able to move the content of `setup.cfg` there.

dsagal:
Thank you!

PR #52, opened 2020-02-12
Fixing issues in my PR.

KOLANICH:


dsagal:
Thanks, but would you mind explaining these changes? Are the new `setup_requires` requirements actually necessary?

alexmojaki:
Some background: I recently [merged in a PR](https://github.com/ipython/ipython/pull/11886) which makes asttokens a dependency of IPython (in git master at least) via [`stack_data`](https://github.com/alexmojaki/stack_data). @KOLANICH saw the new dependencies and wanted to update how they're packaged. They've made similar PRs in my 3 packages which also became dependencies. You can maybe get answers to some of your questions in [one of those PRs here](https://github.com/alexmojaki/stack_data/pull/3).

KOLANICH:
>Are the new setup_requires requirements actually necessary?

Yes.
"setuptools_scm[toml]>=3.4.3" is required because we use newly introduced support for `pyproject.toml` in it.

`setuptools>=44` is needed because earlier versions won't use `pyproject.toml` That's why we duplicate it into `setup.cfg` to make sure earlier versions can use this information.

PR #67, opened 2021-04-07
Drop support for dead pythons

WhyNotHugo:
Drop support for Pythons than reached their end of life and are no longer supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ language: python
 sudo: false
 
 python:
-  - 2.7
-  - 3.5
   - 3.6
   - 3.7
   - 3.8-dev
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -17,8 +17,6 @@ import bisect
 import token
 import tokenize
 import io
-import six
-from six.moves import xrange      # pylint: disable=redefined-builtin
 from .line_numbers import LineNumbers
 from .util import Token, match_token, is_non_coding_token
 from .mark_tokens import MarkTokens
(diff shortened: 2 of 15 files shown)


dsagal:
Sorry but I still need support for 2.7! :)

alexmojaki:
Is there a specific reason to do this? Do you want to make other changes to the codebase afterwards?

Here are some libraries which use asttokens and support 2.7:

- https://github.com/alexmojaki/birdseye
- https://github.com/alexmojaki/snoop
- https://github.com/gruns/icecream

WhyNotHugo:
I'm mostly looking to trim me dependency tree by dropping things like `six` which are only required for compatibility with unsupported versions of Python.

But no, no follow up PR intended, aside from

WhyNotHugo:
Can't those downstream dependees simply depend on an older version of this library, much like they depend on an older Python?

alexmojaki:
> I'm mostly looking to trim me dependency tree by dropping things like six which are only required for compatibility with unsupported versions of Python.

1. I'm curious, where are you using asttokens?
2. Is asttokens your only dependency that uses six? six is quite popular.
3. Would dropping six from your dependency tree really be helpful?
4. Removing six from the dependencies of asttokens looks like it should be quite easy. Only a few symbols are used, it'd be easy to 'implement' them.

> Can't those downstream dependees simply depend on an older version of this library

No, I want them to benefit from future updates to this library.

> much like they depend on an older Python?

They support 2.7, they don't depend on it. They also support 3.9.

WhyNotHugo:
Oh, I see. I guess this PR doesn't make sense in that case.

Again, I've no pressing need to trim `six` immediately, it just bothers me to have dependencies installed only to support software that's long past reached it's end of life.

Sorry for the noise!

PR #68, opened 2021-04-16
setup.cfg: remove wheel dependency

asafkahlon:
This shouldn't be a hard dependency since it's required only
when building binary wheels.

Signed-off-by: Asaf Kahlon <asafka7@gmail.com>

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ classifiers =
 [options]
 packages = asttokens
 install_requires = six
-setup_requires = setuptools>=44; wheel; setuptools_scm[toml]>=3.4.3
+setup_requires = setuptools>=44; setuptools_scm[toml]>=3.4.3
 
 [options.extras_require]
 test = astroid; pytest


alexmojaki:
Hi @asafkahlon, I appreciate the help here and in `executing` as I still don't understand this stuff. Does requiring wheel in setup.cfg mean that users need to have `wheel` to run `pip install asttokens`? Or is this to improve things for developers?

asafkahlon:
Hi @alexmojaki,
Requiring `wheel` on `setup_requires` means the user needs `wheel` installed on his environment.
In your case, the strict dependency on `wheel` will fail the build of the package in case a user tries to run something like `python setup.py build` and doesn't have `wheel` installed.

alexmojaki:
When you say the user *needs* to have `wheel` installed, do they need it for a regular `pip install` or just when running `python setup.py`?

What about `setuptools_scm[toml]`? It's also required by setup.cfg, does it make sense to require that for users?

Both projects have scripts which publish binary wheels to PyPI, it'd be nice if we didn't have to deal with extra errors and manual steps when working from a fresh setup. What should provide that?

What about pyproject.toml? It also specifies `requires = ["setuptools>=44", "wheel", "setuptools_scm[toml]>=3.4.3"]`. When is that used?

alexmojaki:
@KOLANICH you created setup.cfg in these projects, can you give some insight and help me understand? Is @asafkahlon right that `wheel` should be removed?

asafkahlon:
> When you say the user _needs_ to have `wheel` installed, do they need it for a regular `pip install` or just when running `python setup.py`?

It means that `wheel` must be installed when running the `python setup.py ...` command
Generally, the `wheel` package is needed only running `python setup.py bdist_wheel`, and correct me if I'm wrong, it's done only when you want to upload a new version to PyPI.
But, a user may not want to create a `whl` file, just run `python setup.py build` without having `wheel` installed on his environment. At the current state, this can't be done.

> 
> What about `setuptools_scm[toml]`? It's also required by setup.cfg, does it make sense to require that for users?
> 

Regarding `setuptools-scm`, it's needed because you have a section on `pyproject.toml` which declares versioning information.

> Both projects have scripts which publish binary wheels to PyPI, it'd be nice if we didn't have to deal with extra errors and manual steps when working from a fresh setup. What should provide that?
> 
Those scripts should make sure `wheel` is installed before making the `whl` file. According to what I see, it's done on the `travis` file, so you won't need any changes or manual steps when deploying a new version.

> What about pyproject.toml? It also specifies `requires = ["setuptools>=44", "wheel", "setuptools_scm[toml]>=3.4.3"]`. When is that used?

I'm not very familiar with this file, but according to this page it's mainly needed for tools like `pip`: https://packaging.python.org/tutorials/packaging-projects/#creating-pyproject-toml
I think `wheel` can be removed from this file too. I even tried to run `pip install` on the `whl` file without having `wheel` on my environment, and it seems to work.

KOLANICH:
1. >Requiring wheel on setup_requires means the user needs wheel installed on his environment.

It almost always is needed. `setuptools install` is badly broken, it doesn't set entry points, don't rely on it. The only way that works reliably is building a wheel first, then installing it. In fact `pip` works this way when the package is built via `setuptools`. `wheel` is now a de-facto mandatory part of packaging ecosystem relying on `setuptools`. For `poetry` and `flit` `wheel` is not mandatory, they have own impl of what `wheel` does inside of them.

>Those scripts should make sure wheel is installed before making the whl file.

If you have modern enough setuptools, it would install all the deps in pyproject.toml / setup_requires automatically.

>I even tried to run pip install on the whl file without having wheel on my environment, and it seems to work.

`pip` devs have decided to vendor all the deps into `pip`, including `wheel` (so, if you have `pip` and `wheel` installed, you have 2 copues of `wheel`, one standalone and used by `setuptools`, another one is within `pip._vendor`).

alexmojaki:
@asafkahlon why is it useful to be able to run `python setup.py` without having `wheel` present? Why do you want to run `python setup.py build`?

KOLANICH:
I was mistaken, `pip` doesn't vendor `wheel`. It relies on an external `wheel` in the case of `setuptools`.

asafkahlon:
> @asafkahlon why is it useful to be able to run `python setup.py` without having `wheel` present? Why do you want to run `python setup.py build`?

I got to this issue when I was trying to add your package to [Buildroot](https://buildroot.org/).
When creating a root filesystem for an embedded device you probably want to build a package from source, because you may have specific toolchain with tools for your architecture (and the `whl` format is mostly used for having compiled binaries, which may not fit to the device).

alexmojaki:
"compiled binaries" is a bit extreme here, it's basically a zip containing Python source files.

So does Buildroot require users to run `python setup.py`? Is having `wheel` a problem for these users? Were you unable to add the package because of this issue, or did you get it working and this is an optimisation?

What's the bigger story here? I'm guessing this has to do with the icecream package?

asafkahlon:
> "compiled binaries" is a bit extreme here, it's basically a zip containing Python source files.

Yes, I agree your package doesn't contain any compiled binaries, but this is actually the added-value of using `whl`.

> 
> So does Buildroot require users to run `python setup.py`? Is having `wheel` a problem for these users? Were you unable to add the package because of this issue, or did you get it working and this is an optimisation?

No, this is not an optimization. The package can't be added to Buildroot for now because of the `wheel` dependency.
By the way, this is not the first time we encounter this issue, and there were some commits on other repositories with the same change. One example: https://github.com/PyO3/setuptools-rust/commit/1d7ee9fc239c846fdd5aac5f3a28705333d47d77

> 
> What's the bigger story here? I'm guessing this has to do with the icecream package?

Yes, what I really want is to be able to work with `icecream` on my device and add it to Buildroot. For that, I need to add all of `icecream` dependencies (which don't exist on Buildroot).

alexmojaki:
Can you provide any other similar examples?

Have you been able to use icecream on your device? If not, can you try using either one of the single file builds in https://github.com/gruns/icecream/tree/9e1bc283d7b263ea432ce673fd05415479fa7d07/single_file (e.g. icecream_py39.py) or https://github.com/salabim/ycecream? I'm interested to see how well asttokens, executing, and pygments work in a constrained environment.

asafkahlon:
> Can you provide any other similar examples?
> 

This is an example of a similar commit: https://github.com/pyca/pynacl/commit/44fb453931c611d4822343debe902cca8e1c8afe

> Have you been able to use icecream on your device? If not, can you try using either one of the single file builds in https://github.com/gruns/icecream/tree/9e1bc283d7b263ea432ce673fd05415479fa7d07/single_file (e.g. icecream_py39.py) or https://github.com/salabim/ycecream? I'm interested to see how well asttokens, executing, and pygments work in a constrained environment.

I didn't work with `icecream` on my device because I couldn't build it...
I can give a try to the single-file alternative, but I think it's another issue.

dsagal:
I don't have much to contribute to this discussion, since I am no expert with Python packaging systems. I can say that I see both points. Alex's comment has truth to it:

> Both projects have scripts which publish binary wheels to PyPI, it'd be nice if we didn't have to deal with extra errors and manual steps when working from a fresh setup. What should provide that?

With this change, preparing to publish (for which I run `python setup.py sdist bdist_wheel`) requires wheel to be available. Just doing `pip install -e '.[test]'` in a clean environment will no longer bring in wheel, so installing it would be an extra step (a simple one, `pip install wheel`).

That said, wheel isn't needed to *use* `asttokens`, or even to contribute to it. In other words, the extra annoyance from removing the wheel dependency seems minor and would affect a select few people.

So I have no objections to this PR if it helps others. I'd like @alexmojaki or @KOLANICH to OK it though, in case they see is a more serious tradeoff than I realize.

alexmojaki:
> I don't have much to contribute to this discussion, since I am no expert with Python packaging systems. I can say that I see both points. Alex's comment has truth to it:
> 
> > Both projects have scripts which publish binary wheels to PyPI, it'd be nice if we didn't have to deal with extra errors and manual steps when working from a fresh setup. What should provide that?
> 
> With this change, preparing to publish (for which I run `python setup.py sdist bdist_wheel`) requires wheel to be available. Just doing `pip install -e '.[test]'` in a clean environment will no longer bring in wheel, so installing it would be an extra step (a simple one, `pip install wheel`).

This was a side comment, it's the least of my concerns. `wheel` could be added to `test` or some other extra. `twine` isn't automatically installed anywhere either and it's also needed to publish.

The problem is that I also don't understand this stuff, everyone has different requirements, things are constantly changing, and I'm scared of somehow breaking things for other people. So I'm very wary of touching things.

But I think @asafkahlon has provided enough evidence that this is safe and useful, so I've merged the similar PR https://github.com/alexmojaki/executing/pull/21 and released, and nothing has broken *yet*. I think this is OK.

KOLANICH (approved):

PR #72, opened 2021-09-05
Typing support

palfrey:


asttokens/util.py
```diff
@@ -59,21 +68,24 @@ def expect_token(token, tok_type, tok_str=None):
 
 # These were previously defined in tokenize.py and distinguishable by being greater than
 # token.N_TOKEN. As of python3.7, they are in token.py, and we check for them explicitly.
-if hasattr(token, 'ENCODING'):
+if sys.version_info >= (3, 7):
```
palfrey, line 71: Because of https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks

tests/test_asttokens.py
```diff
@@ -80,6 +80,8 @@ def test_token_methods(self):
     self.assertTrue(token.ISEOF(atok.find_token(atok.tokens[5], tokenize.NL).type))
 
   def test_unicode_offsets(self):
+    if six.PY2:
+      pytest.skip("Unicode parsing is different on Python 2")
```
palfrey, line 84: No idea how anyone got this working on 2.7. This should have failed before now! See also https://github.com/gristlabs/asttokens/runs/4493079453?check_suite_focus=true

tests/test_mark_tokens.py
```diff
@@ -560,6 +561,8 @@ def test_statements_with_semicolons(self):
 
 
   def test_complex_numbers(self):
+    if six.PY2:
+      pytest.skip("Complex number parsing is different on Python 2")
```
palfrey, line 565: Similar 2.7 failure to the one above

palfrey:
Any thoughts on this @dsagal @alexmojaki?

asttokens/asttokens.py
```diff
@@ -43,14 +43,14 @@ class ASTTokens(object):
...
+    # type: (str, bool, Optional[Module], str) -> None
     self._filename = filename
     self._tree = ast.parse(source_text, filename) if parse else tree
 
     # Decode source after parsing to let Python 2 handle coding declarations.
     # (If the encoding was not utf-8 compatible, then even if it parses correctly,
     # we'll fail with a unicode error here.)
-    if isinstance(source_text, six.binary_type):
-      source_text = source_text.decode('utf8')
+    source_text = six.ensure_str(source_text)
```
alexmojaki, line 53: Should probably be `ensure_text`

asttokens/asttokens.py
```diff
@@ -66,56 +66,65 @@ def __init__(self, source_text, parse=False, tree=None, filename='<unknown>'):
...
 
   def _generate_tokens(self, text):
+    # type: (str) -> Iterator[Token]
     """
     Generates tokens for the given code.
     """
     # This is technically an undocumented API for Python3, but allows us to use the same API as for
     # Python2. See http://stackoverflow.com/a/4952291/328565.
-    for index, tok in enumerate(tokenize.generate_tokens(io.StringIO(text).readline)):
+    for index, tok in enumerate(tokenize.generate_tokens(six.StringIO(text).readline)):
```
alexmojaki, line 88: Since you're having Python 2 unicode issues, can you revert this and see if it helps?

asttokens/asttokens.py
```diff
@@ -175,13 +193,15 @@ def token_range(self, first_token, last_token, include_extra=False):
         yield self._tokens[i]
 
   def get_tokens(self, node, include_extra=False):
+    # type: (NodeNG, bool) -> Iterator
```
alexmojaki, line 196:
```suggestion
    # type: (NodeNG, bool) -> Iterator[Token]
```

asttokens/mark_tokens.py
```diff
@@ -147,11 +157,13 @@ def _expand_to_matching_pairs(self, first_token, last_token, node):
   # that will actually be assigned.
 
   def visit_default(self, node, first_token, last_token):
+    # type: (NodeNG, util.Token, util.Token) -> Tuple[util.Token, util.Token]
     # pylint: disable=no-self-use
     # By default, we don't need to adjust the token we computed earlier.
     return (first_token, last_token)
 
   def handle_comp(self, open_brace, node, first_token, last_token):
+    # type: (NodeNG, Optional[str], util.Token, util.Token) -> Tuple[util.Token, util.Token]
```
alexmojaki, line 166:
```suggestion
    # type: (str, NodeNG, util.Token, util.Token) -> Tuple[util.Token, util.Token]
```

asttokens/mark_tokens.py
```diff
@@ -162,28 +174,38 @@ def handle_comp(self, open_brace, node, first_token, last_token):
...
+      # type: (NodeNG, util.Token, util.Token) -> Tuple[util.Token, util.Token]
       return self.handle_comp('{', node, first_token, last_token)
 
     def visit_dictcomp(self, node, first_token, last_token):
+      # type: (NodeNG, util.Token, util.Token) -> Tuple[util.Token, util.Token]
       return self.handle_comp('{', node, first_token, last_token)
 
-  def visit_comprehension(self, node, first_token, last_token):
+  def visit_comprehension(self,
+                          node,  # type: util.Token
```
alexmojaki, line 191:
```suggestion
                          node,  # type: NodeNG
```

Is it possible to make a type alias to apply to all these methods?

asttokens/mark_tokens.py
```diff
@@ -232,11 +257,17 @@ def visit_call(self, node, first_token, last_token):
       first_token = self._code.next_token(first_token)
     return (first_token, last_token)
 
-  def visit_subscript(self, node, first_token, last_token):
+  def visit_subscript(self,
+                      node,  # type: util.Token
```
alexmojaki, line 261:
```suggestion
                      node,  # type: NodeNG
```

asttokens/mark_tokens.py
```diff
@@ -232,11 +257,17 @@ def visit_call(self, node, first_token, last_token):
...
+                      first_token,  # type: util.Token
+                      last_token,  # type: util.Token
+                      ):
+    # type: (...) -> Tuple[util.Token, util.Token]
     last_token = self.handle_following_brackets(node, last_token, '[')
     return (first_token, last_token)
 
   def visit_slice(self, node, first_token, last_token):
+    # type: (util.Token, util.Token, util.Token) -> Tuple[util.Token, util.Token]
```
alexmojaki, line 270:
```suggestion
    # type: (NodeNG, util.Token, util.Token) -> Tuple[util.Token, util.Token]
```

asttokens/mark_tokens.py
```diff
@@ -256,6 +287,7 @@ def visit_slice(self, node, first_token, last_token):
     return (first_token, last_token)
 
   def handle_bare_tuple(self, node, first_token, last_token):
+    # type: (util.Token, util.Token, util.Token) -> Tuple[util.Token, util.Token]
```
alexmojaki, line 290:
```suggestion
    # type: (NodeNG, util.Token, util.Token) -> Tuple[util.Token, util.Token]
```

asttokens/util.py
```diff
@@ -201,12 +222,13 @@ def visit_tree(node, previsit, postvisit):
...
         stack.insert(ins, (n, pvalue, _PREVISIT))
     else:
-      ret = postvisit(current, par_value, value)
+      ret = postvisit(current, par_value, cast(Optional[Token], value))
   return ret
 
 
 
 def walk(node):
+  # type: (Module) -> Iterator[Module]
```
alexmojaki, line 231:
```suggestion
  # type: (Module) -> Iterator[NodeNG]
```

asttokens/mark_tokens.py
palfrey: Maybe. Not sure how!

asttokens/util.py
palfrey: I don't think this is right, given the first item it returns is the input Module. Possibly `Union[Module, NodeNG]` but hard to tell in the full-on dynamic typing going on here.

asttokens/asttokens.py
palfrey: Tried this, and it breaks the Python 2 typing

asttokens/util.py
alexmojaki: Most of the mentions of `NodeNG` aren't right since the functions work with either an `astroid` or `ast` node. I thought you were just using `NodeNG` to mean `ast.AST` for some reason. But this definitely yields more than just modules.

alexmojaki:
Can you revert the skips you added for Python 2 and see if they pass now?

palfrey:
> Can you revert the skips you added for Python 2 and see if they pass now?

I've just tried that locally, and they're still broken

palfrey:
> Please revert the skips and the unicode related changes and push. The tests were working before, and only adding type comments shouldn't change that. If the previous code broke typing, that's a typing problem and that's secondary.

Yup, you're right. I reset bits of the changes, removed the test skips and added a typing workaround [for an upstream bug](https://github.com/python/typeshed/issues/7003) and it's working better now. Still one test to fix, but working on it...

asttokens/asttokens.py
palfrey: Resolved. [Upstream typing issue](https://github.com/python/typeshed/issues/7003) and there's now a `cast` to fix the type as a workaround.

asttokens/asttokens.py
palfrey: Yup, you were right. Done now.

asttokens/mark_tokens.py
```diff
@@ -331,11 +381,13 @@ def handle_num(self, node, value, first_token, last_token):
     return (first_token, last_token)
 
   def visit_num(self, node, first_token, last_token):
+    # type: (NodeNG, util.Token, util.Token) -> Tuple[util.Token, util.Token]
     return self.handle_num(node, node.n, first_token, last_token)
 
   # In Astroid, the Num and Str nodes are replaced by Const.
   def visit_const(self, node, first_token, last_token):
-    if isinstance(node.value, numbers.Number):
```
alexmojaki, line 338: This is probably the source of the complex number test failure...

palfrey: And solved by c0bdb710fe790b73bcad3cef4629119fb075b989

asttokens/util.py
palfrey: I've just ran some experiments on this one and found the correct set of types was `Union[Module, AST, NodeNG]`

Are you saying that most of the `NodeNG` refs should actually be `Union[AST, NodeNG]`?

alexmojaki: Sorry I forgot about this. Yes, technically, I think they should usually be `Union[AST, NodeNG]`, i.e. they work with both `ast` and `astroid`.

palfrey: I've added a `AstNode` type alias for `Union[AST, NodeNG]` and moved over most of the util usages of `NodeNG` to that. There are two exceptions, because they use features only on `NodeNG`

asttokens/util.py
```diff
@@ -124,30 +140,36 @@ def iter_children_ast(node):
...
 
 
 def is_starred(node):
+  # type: (AstNode) -> bool
   """Returns whether node is a starred expression node."""
   return node.__class__.__name__ == 'Starred'
 
 
 def is_slice(node):
+  # type: (NodeNG) -> bool
```
palfrey, line 172: `.elts` is only on `NodeNG`, so can't use `ASTNode`

alexmojaki: It definitely exists in `ast`:

```
>>> ast.parse('1, 2').body[0].value.elts
[<_ast.Constant object at 0x7f56dd8ba5b0>, <_ast.Constant object at 0x7f56dd8ba3d0>]
```

Maybe you need to specify specific types of node?

setup.cfg
```diff
@@ -36,7 +36,13 @@ install_requires = six
 setup_requires = setuptools>=44; setuptools_scm[toml]>=3.4.3
 
 [options.extras_require]
-test = astroid<=2.5.3; pytest
+test = 
+    astroid<=2.5.3
+    pytest
+    typing; python_version < "3.5"
```
alexmojaki, line 42: Isn't this an essential (i.e. not extra) requirement for Python 2 users?

asttokens/mark_tokens.py
```diff
@@ -61,14 +68,15 @@ def _visit_before_children(self, node, parent_token):
     return (token or parent_token, token)
 
   def _visit_after_children(self, node, parent_token, token):
+    # type: (NodeNG, Optional[util.Token], Optional[util.Token]) -> None
```
alexmojaki, line 71: There's still lots of `NodeNG` here and in `asttokens.py` that should really be `AstNode`. Note that the library adds several attributes like `first_token` to the nodes, you'll have to find a way to deal with that.

setup.cfg
palfrey: Yup! Fixed

asttokens/mark_tokens.py
palfrey: I've replaced them with AstNode, and fixed the definition of that to use an `EnhancedAST` types marker class that basically just goes "we promise ASTs have some extra fields"

asttokens/asttokens.py
```diff
@@ -175,13 +198,15 @@ def token_range(self, first_token, last_token, include_extra=False):
         yield self._tokens[i]
 
   def get_tokens(self, node, include_extra=False):
+    # type: (AstNode, bool) -> Iterator[Token]
     """
     Yields all tokens making up the given node. If include_extra is True, includes non-coding
     tokens such as tokenize.NL and .COMMENT.
     """
-    return self.token_range(node.first_token, node.last_token, include_extra=include_extra)
+    return self.token_range(cast(Token, node.first_token), cast(Token, node.last_token), include_extra=include_extra)
```
alexmojaki, line 206: Doesn't `EnhancedAST` remove the need for these casts?

asttokens/mark_tokens.py
```diff
@@ -195,9 +219,10 @@ def handle_attr(self, node, first_token, last_token):
   visit_delattr = handle_attr
 
   def handle_def(self, node, first_token, last_token):
+    # type: (AstNode, util.Token, util.Token) -> Tuple[util.Token, util.Token]
     # With astroid, nodes that start with a doc-string can have an empty body, in which case we
     # need to adjust the last token to include the doc string.
-    if not node.body and getattr(node, 'doc', None):
+    if not getattr(node, 'body', None) and getattr(node, 'doc', None):
```
alexmojaki, line 225: Why this change?

palfrey: This was mostly because it predated the work with narrowing the types for the various handle/visit functions down to the ones they were actually using. I tried doing that here, and the list of types is pretty large (I got to about at least 6 triggered by the tests), and they're annoyingly lacking in common parent classes to use instead. I've reset this back to it's original code and added a `type: ignore` comment.

asttokens/asttokens.py
palfrey: Yup, removed.

alexmojaki:
:+1: I'm happy. Thanks for your patience! @dsagal ?

alexmojaki:
Thanks for all your hard work @palfrey !

PR #74, opened 2021-09-18
Correctly mark slice token spans.

ssbr:
Previous to this, the slice `x[1:]` would have the token span for `1`, instead of `1:`.

This is not the ideal way to do it -- I think ideally, mark_tokens should be using both `col_offset` and `end_col_offset`, in which case this would have worked automatically on Python 3.9. But this does get it working "as much as possible" in versions prior to 3.9.

In particular: while slices like `1:` and `:1` will now work correctly in Python 3.8-, the slice `x[:]` will still get attributed to `x` and not `:`. Oops. That issue does not exist in 3.9 (before or after this change).

### DIFF
--- a/asttokens/mark_tokens.py
+++ b/asttokens/mark_tokens.py
@@ -236,6 +236,25 @@ class MarkTokens(object):
     last_token = self.handle_following_brackets(node, last_token, '[')
     return (first_token, last_token)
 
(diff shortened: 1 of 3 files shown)


dsagal (approved):
Looks good to me, thank you!

Issue #73, opened 2021-09-05
Fix travis build


palfrey:
https://travis-ci.org/github/gristlabs/asttokens is now marked as not working any more since their migration, and  https://app.travis-ci.com/github/gristlabs/asttokens looks vaguely active, but not actually running anything. The latter probably needs switching on?

PR #75, opened 2021-09-30
Move testing from Travis to GitHub Actions

dsagal:
Resolves https://github.com/gristlabs/asttokens/issues/73. Couldn't get travis-ci.com to work, but Github Actions work well.


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
+name: Build and test
+on: [push]
```
palfrey, line 2: This won't build on PRs

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - 2.7
+          - 3.5
+          - 3.6
+          - 3.7
+          - 3.8
+          - 3.9
```
alexmojaki, line 15: 3.10 final is being released in a few days...

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
...
+          - pypy2
+          - pypy3
+    env:
+      # ASTTOKENS_SLOW_TESTS: 1
```
alexmojaki, line 19: Is this intentional?

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Install dependencies
+        run: |
+          pip install --upgrade setuptools>=44 wheel setuptools_scm[toml]>=3.4.3 pep517
+          pip install --upgrade coverage
+          # python3 -m pep517.build . # replace the following command with this one when you drop python 2
+          python ./setup.py bdist_wheel
+          export WHLNAME=./dist/asttokens-0.CI-py2.py3-none-any.whl
+          mv ./dist/*.whl $WHLNAME
+          pip install --upgrade --pre "$WHLNAME[test]"
```
alexmojaki, line 38: I have a similar packaging setup in `executing` and Python 2 support, and I just use `pip install .` instead of all this: https://github.com/alexmojaki/executing/blob/master/.github/workflows/test.yml

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+
+      - name: Coveralls Finished
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          pip3 install --upgrade coveralls
+          python3 -m coveralls --service=github --finish
```
alexmojaki, line 64:
```suggestion
      - name: Coveralls Finished
        uses: AndreMiras/coveralls-python-action@v20201129
        with:
          parallel-finished: true
```

Also based on https://github.com/alexmojaki/executing/blob/master/.github/workflows/test.yml

.github/workflows/build-and-test.yml
alexmojaki:
```suggestion
          - 3.9
          - 3.10-dev
```

.github/workflows/build-and-test.yml
palfrey:
```suggestion
on:
  push:
    branches:
      - master
  pull_requests:
```

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,68 @@
...
+          # TODO Remove the commented-out lines if it works.
+          # pip install --upgrade setuptools>=44 wheel setuptools_scm[toml]>=3.4.3 pep517
+          # pip install --upgrade coverage
+          # # python3 -m pep517.build . # replace the following command with this one when you drop python 2
+          # python ./setup.py bdist_wheel
+          # export WHLNAME=./dist/asttokens-0.CI-py2.py3-none-any.whl
+          # mv ./dist/*.whl $WHLNAME
+          # pip install --upgrade --pre "$WHLNAME[test]"
+          pip install --upgrade coveralls pytest setuptools setuptools_scm pep517
+          pip install .
```
alexmojaki, line 47:
```suggestion
          pip install .[test]
```

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,68 @@
...
+      fail-fast: false
+      matrix:
+        python-version:
+          - 2.7
+          - 3.5
+          - 3.6
+          - 3.7
+          - 3.8
+          - 3.9
+          - 3.10
```
alexmojaki, line 20:
```suggestion
          - '3.10'
```

.github/workflows/build-and-test.yml
alexmojaki:
```suggestion
      ASTTOKENS_SLOW_TESTS: 1
```

dsagal:
Thank you @alexmojaki ! This is great. I'd only prefer to disable the "slow tests" -- the tests take many times longer, but the benefit is small, and there is virtually no difference in coverage. So it doesn't seem worth enabling.

alexmojaki:
Coverage isn't the point, the slow tests uncovered bugs back when they were added. They test things we wouldn't know need testing. I think it's fine for CI to be slow as long as you can test quickly locally, so CI is just to double check. If it's disabled here, when will the extra tests run? I suggest either:

- Test a random sample of modules up to a time limit.
- Test slow tests on push to master, but not pull requests.
- Split fast and slow tests into separate checks, so when github says 9 checks have passed you know the fast tests have passed.

On a related note, there should probably be some tests of the new pattern matching syntax in 3.10.

dsagal:
OK, I added it back because I want any tests rather than no tests. But I think fast targeted tests are better than running over all python code we can get our hands on just in the hopes that it includes some rare constructs that should be tested but aren't.

alexmojaki:
Of course it's better to directly test all the cases we know about, but the sys.modules tests are good at filling in the gaps. Here's when these tests were added: https://github.com/gristlabs/asttokens/pull/41

Here are commits from that PR representing fixes for previously unknown bugs that were only found by the new tests:

- https://github.com/gristlabs/asttokens/pull/41/commits/46885eec85857f3fc3ec1b08c8329d43199b338b - this case led to a fix in Python itself: https://bugs.python.org/issue38535
- https://github.com/gristlabs/asttokens/pull/41/commits/879ea47be89da05566096589e6fcabf3b6771adb
- https://github.com/gristlabs/asttokens/pull/41/commits/e0afa320f1424c1fc99666ec1759522f11bf46d6
- https://github.com/gristlabs/asttokens/pull/41/commits/8bde449501caa05dc5db21f564047b87f6c10c79
- https://github.com/gristlabs/asttokens/pull/41/commits/0b18bd0a13a71ba692991aff2e11a926bbe868ae
- https://github.com/gristlabs/asttokens/pull/41/commits/b78fc6a62d20adde5aee4464e05a89ce071dae3f

alexmojaki (approved):

PR #78, opened 2022-03-26
Upgrade pypy test list to all the available ones

palfrey:
The old form is getting removed as of [setup-python v3](https://github.com/actions/setup-python/releases/tag/v3.0.0), and it's worth probably testing all of them.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -18,8 +18,11 @@ jobs:
           - 3.8
           - 3.9
           - '3.10'
-          - pypy2
-          - pypy3
+          # As per https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#pypy list of versions
+          - pypy-2.7
+          - pypy-3.6
+          - pypy-3.7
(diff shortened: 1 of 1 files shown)


alexmojaki:
Sounds good to me. @dsagal ?

palfrey:
@alexmojaki Can this be merged?

alexmojaki:
@palfrey it looks like the combination of this and #72 don't work. The last build on master failed for every version on the mypy step: https://github.com/gristlabs/asttokens/actions/runs/2141385979

palfrey:
https://github.com/gristlabs/asttokens/blob/master/.github/workflows/build-and-test.yml#L48 needs editing for the new build names. I'd expected 3.8 to work, but apparently not, so that needs looking at

palfrey:
Sadly, pypy 3.8's promises of working with mypy were wrong as there's [still at least one bug](https://foss.heptapod.net/pypy/pypy/-/issues/3526). https://github.com/gristlabs/asttokens/pull/80 filters out all those builds.

TBH, I was expecting you to merge one of the PRs, update the other to master and see failures, not just merge both!

PR #82, opened 2022-07-31
Workaround Python tokenize bug with non-ASCII characters

alexmojaki:
Workaround for https://github.com/python/cpython/issues/68382

I ran into this recently in futurecoder and implemented the core logic in https://github.com/alexmojaki/futurecoder/pull/373/commits/ea8bed30f9eaa795179b5beebe2e026ee53ee0e8#diff-2439d77444f0be435d92d3e5df78ab2f065592d4f66c075a1bc92e1b8bb41954R25-R46 to support variable names translated to Tamil. But monkeypatching `tokenize` globally leads to incorrect behaviour with actually invalid code, which for example causes friendly-traceback to explain syntax errors incorrectly. Besides, this fix belongs here.

### DIFF
--- a/asttokens/util.py
+++ b/asttokens/util.py
@@ -332,23 +332,33 @@ else:
     Workaround for https://github.com/python/cpython/issues/68382.
     Should only be used when tokenizing a string that is known to be valid syntax,
     because it assumes that error tokens are not actually errors.
-    Combines groups of consecutive NAME and/or ERRORTOKEN tokens into a single NAME token.
+    Combines groups of consecutive NAME, NUMBER, and/or ERRORTOKEN tokens into a single NAME token.
     """
-    for key, group_iter in itertools.groupby(
-        original_tokens,
-        lambda t: tokenize.NAME if t.type == tokenize.ERRORTOKEN else t.type,
-    ):
-      group = list(group_iter)  # type: List[tokenize.TokenInfo]
-      if key == tokenize.NAME and len(group) > 1 and any(tok.type == tokenize.ERRORTOKEN for tok in group):
-        line = group[0].line  # type: str
-        assert {tok.line for tok in group} == {line}
-        yield tokenize.TokenInfo(
-          type=tokenize.NAME,
-          string="".join(t.string for t in group),
-          start=group[0].start,
-          end=group[-1].end,
-          line=line,
-        )
+    group = []  # type: List[tokenize.TokenInfo]
+    for tok in original_tokens:
+      if (
+          tok.type in (tokenize.NAME, tokenize.ERRORTOKEN, tokenize.NUMBER)
+          # Only combine tokens if they have no whitespace in between
+          and (not group or group[-1].end == tok.start)
+      ):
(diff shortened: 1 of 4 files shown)


asttokens/util.py
```diff
@@ -311,3 +316,39 @@ def get(self, obj, cls):
...
+    Combines groups of consecutive NAME and/or ERRORTOKEN tokens into a single NAME token.
+    """
+    for key, group_iter in itertools.groupby(
+        original_tokens,
+        lambda t: tokenize.NAME if t.type == tokenize.ERRORTOKEN else t.type,
+    ):
+      group = list(group_iter)  # type: List[tokenize.TokenInfo]
+      if key == tokenize.NAME and len(group) > 1 and any(tok.type == tokenize.ERRORTOKEN for tok in group):
+        line = group[0].line  # type: str
+        assert {tok.line for tok in group} == {line}
```
dsagal, line 344: Should this check that the end of one token is the start of the next one in this group, in order to avoid merging tokens that are separated by whitespace?

asttokens/util.py
```diff
@@ -311,3 +316,39 @@ def get(self, obj, cls):
...
+    Should only be used when tokenizing a string that is known to be valid syntax,
+    because it assumes that error tokens are not actually errors.
+    Combines groups of consecutive NAME and/or ERRORTOKEN tokens into a single NAME token.
+    """
+    for key, group_iter in itertools.groupby(
+        original_tokens,
+        lambda t: tokenize.NAME if t.type == tokenize.ERRORTOKEN else t.type,
+    ):
+      group = list(group_iter)  # type: List[tokenize.TokenInfo]
+      if key == tokenize.NAME and len(group) > 1 and any(tok.type == tokenize.ERRORTOKEN for tok in group):
```
dsagal, line 342: What about a name like `℘·12` (like the example, but with a numeric suffix)? It would involve a token of type `NUMBER` too

alexmojaki: You're right, that breaks this. So I don't think `itertools.groupby` works well here. I think I have to accumulate each group manually, such that consecutive tokens that have no whitespace between them and are all of type NAME/NUMBER/ERRORTOKEN are in the same group.

asttokens/util.py
```diff
@@ -332,23 +332,33 @@ def patched_generate_tokens(original_tokens):
...
+          # Only combine tokens if they have no whitespace in between
+          and (not group or group[-1].end == tok.start)
+      ):
+        group.append(tok)
       else:
-        for tok in group:
-          yield tok
+        for combined_token in combine_tokens(group):
+          yield combined_token
+        group = [tok]
```
dsagal, line 348: Is this logic correct when a token other than (NAME, ERRORTOKEN, NUMBER) is immediately followed by a (NAME, ERRORTOKEN, NUMBER) token? It seems to me like this could unintentionally combine unrelated tokens, as long as there is no whitespace in-between and one of them is ERRORTOKEN.

dsagal (approved):
Thanks!

PR #83, opened 2022-08-01
Allow triggering GitHub actions workflow manually

alexmojaki:
https://github.com/gristlabs/asttokens/pull/82 ran into some extreme corner case (I don't know the details yet) and I wanted to check if the same failure also happened on master. According to https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow, something like this is needed to enable what I wanted to do.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - master
   pull_request:
+  workflow_dispatch:
 jobs:
   test:
     runs-on: ubuntu-latest


alexmojaki:
So yes, the same 3.7 astroid failure also happens on master, and apparently there's also new failures that are specifically fixed by #82. I guess some library code has changed over time to make these surface. I don't think it's the behaviour of Python or dependencies that's changed, just the contents of files tested by `test_sys_modules`.

dsagal (approved):

Issue #89, opened 2022-08-13
mypy reports `error: Name "asttokens.ASTTokens" is not defined`


mristin:
Hi!
I updated to the latest version of asttokens 2.0.7 on my package icontract during a CI run, and the mypy started complaining that ``asttokenst.ASTTokens`` is missing. 

Please see this error: https://github.com/Parquery/icontract/runs/7819826733?check_suite_focus=true

Here's a relevant excerpt:

```
Collecting mypy==0.950
Collecting asttokens<3,>=2
  Downloading asttokens-2.0.7-py2.py3-none-any.whl (23 kB)

...

icontract/_represent.py:41: error: Name "asttokens.ASTTokens" is not defined
icontract/_represent.py:188: error: Name "asttokens.ASTTokens" is not defined
icontract/_represent.py:210: error: Name "asttokens.ASTTokens" is not defined
icontract/_represent.py:264: error: Module has no attribute "ASTTokens"
```

I fixed the issue by importing ``asttokens.asttokens`` and writing out ``asttokens.asttokens.ASTTokens`` instead of importing ``asttokens`` and using ``asttokens.ASTTokens``. Please see this pull request: https://github.com/Parquery/icontract/pull/252

I tried to figure out what caused this mypy error. Honestly, I couldn't really understand where the culprit is. I installed the latest mypy version (0.971) and that didn't help either. I see that you haven't changed ``asttokens/__init__.py`` in 6 years so that can't be a reason either.

Here is a minimal code snippet if you want to reproduce the issue (I called the file ``deleteme.py``):

```
import asttokens

_ = asttokens.ASTTokens("True", parse=True)
```

Then, after installing mypy==0.971 and asttokens==2.0.7, run:

```
mypy --strict deleteme.py
```

I get the error message:

```
deleteme.py:3: error: Module has no attribute "ASTTokens"
Found 1 error in 1 file (checked 1 source file)
```

alexmojaki:
Type hints in the package have only just been released, it's been a bit bumpy.

@palfrey any idea what might cause this?

palfrey:
This appears to be an issue with the strict flag, and `mypy --strict --implicit-reexport deleteme.py` works. See https://stackoverflow.com/questions/60734086/mypy-gives-an-error-while-importing-submodule-module-has-no-attribute and https://github.com/python/mypy/issues/7030

It's fixable in the source code, and I'll have a look at that now.

palfrey:
[...]

PR #90, opened 2022-08-14
Fix mypy explicit re-export issues

palfrey:
Fixes https://github.com/gristlabs/asttokens/issues/89. The easiest way to get an example that exercised enough things was to re-use the test code. It's got _very_ loose typing settings for the moment, as it doesn't really need that much to reproduce the error in this case and making the tests fully typed would be a lot more effort!


alexmojaki:
Thanks! I'll release this soon.

Issue #6, opened 2017-06-06
Nodes that come from f-string parents lack token information


LiraNuna:
Because of https://bugs.python.org/issue29051, a temporary solution (See https://github.com/gristlabs/asttokens/issues/4) was placed to avoid issues.

This issue is to track this workaround's life time so it could be fixed once the above python issue is fixed.

jamescooke:
@LiraNuna thanks for keeping this issue open and for the explanations in #4 about what's going on :+1: 

The original Python issue 29051 was split into two - and the first one https://bugs.python.org/issue30465 "FormattedValue expressions have wrong lineno and col_offset information" is now resolved. Does this mean that this bug can be solved properly now?

For reference, I'm about to wire in a work around to fix this here: https://github.com/jamescooke/flake8-aaa/issues/101 - probably something to prevent the error being raised when the linter tries to access the token information on an f-string.

LiraNuna:
It's been a while and I have to admit I forgot most of the details.

alexmojaki:
I'm not really up for a PR right now, but here's an idea for an algorithm:

1. Get the range of tokens (string literals) representing the JoinedString
2. For each FormattedValue (iterating through the string tokens in parallel):
    1. Find the starting `{` in the current string token.
    2. Determine which character is the end of the main expression: `}`, `!`, or `:`, based on the conversion and format specifier values in the node.
    3. For each such character after the starting position:
        1. Try parsing the text between start and end.
        2. If successful, compare the resulting tree to the FormattedValue tree.
        3. If equal, mark tokens on the subtree, then readjust positions based on the starting position of the string.
    4. Repeat for format specifier if present.

alexmojaki:
The new `asttokens.ASTText` class now supports nodes inside f-strings for Python 3.9.7+ with the `.get_text*` methods. But there's still no token information.

PR #92, opened 2022-09-04
Versions of get_text and get_text_range that don't need mark_tokens

alexmojaki:
As mentioned in https://github.com/ipython/ipython/issues/13731#issuecomment-1227272665, although I think keeping the functionality here is still best, even if tokens aren't usually used.

For now I've just added two methods to `ASTTokens` so that they can be tested and compared to the original methods. Here's a proposal for an actual API:

1. Add a boolean argument `mark` (or `mark_tokens`) to the `ASTTokens` constructor.
2. If `mark=False`, then don't call `self.mark_tokens`.
3. `mark=True` by default for backwards compatibility. Personally I've often used the `first_token` and `last_token` attributes added to AST nodes.
4. `mark=False` is ignored for Python versions older than 3.8. It's still allowed so that callers don't have to check the Python version themselves. Same for PyPy, which apparently doesn't have correct position info either, even in 3.8.
5. `mark=False` is an error for an astroid tree. Even if it's easy to get it to work, I'm personally not interested in maintaining that.
6. The usual `get_text` and `get_text_range` methods should automatically pick the correct implementation based on whether the nodes have been marked with tokens. Then code that only uses these methods only needs to add `mark=False` to the constructor.
7. For cases where the unmarked implementation has known bugs, use the original implementation, calling `mark_tokens` first if necessary. Currently, this means `arg` and `Starred` nodes for Python 3.8, i.e. there's no problem for 3.9+. Those node types only seem to fail under specific conditions which I think are known, but for safety we can just check the type.
8. Ideally this should just start working automatically on nodes inside f-strings. These currently don't work here (#6) but `ast.get_source_segment` does, so the new implementation should too, hopefully transparently.
9. Add new method(s) to replace using `node.first_token.start` and `node.last_token.end`, i.e. they should return (lineno, col_offset) pairs instead of string indices. These will usually be the same as the standard `ast` values but not always.

### DIFF
--- a/tests/tools.py
+++ b/tests/tools.py
@@ -143,20 +143,22 @@ class MarkChecker(object):
         with test_case.assertRaises(SyntaxError, msg=(text, ast.dump(node))):
           test_case.parse_snippet(text, node)
 
-  # Node types that check_get_text_unmarked should ignore.
-  bad_unmarked_types = (
-    # get_text_unmarked does something sensible for modules, but it differs from get_text.
-    ast.Module,
-    # no lineno
-    ast.arguments, ast.withitem,
-  )  # type: Tuple[Type[ast.AST], ...]
-  if sys.version_info[:2] == (3, 8):
-    bad_unmarked_types += (
-      # get_text_unmarked works incorrectly for these types due to bugs in Python 3.8.
-      ast.arg, ast.Starred,
-      # no lineno in 3.8
-      ast.Slice, ast.ExtSlice, ast.Index, ast.keyword,
+  # Node types that check_get_text_unmarked should ignore. Only relevant for Python 3.8+.
+  bad_unmarked_types = ()  # type: Tuple[Type[ast.AST], ...]
+  if sys.version_info[:2] >= (3, 8):
+    bad_unmarked_types = (
+      # get_text_unmarked does something sensible for modules, but it differs from get_text.
+      ast.Module,
+      # no lineno
+      ast.arguments, ast.withitem,
     )
+    if sys.version_info[:2] == (3, 8):
+      bad_unmarked_types += (
+        # get_text_unmarked works incorrectly for these types due to bugs in Python 3.8.
+        ast.arg, ast.Starred,
+        # no lineno in 3.8
+        ast.Slice, ast.ExtSlice, ast.Index, ast.keyword,
+      )
 
   def _check_alias_unmarked(self, node, test_case, text_unmarked):
     if sys.version_info < (3, 10):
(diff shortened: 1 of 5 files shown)


tests/tools.py
```diff
@@ -111,6 +114,47 @@ def verify_all_nodes(self, test_case):
...
+        or 'pypy' in sys.version.lower()
+    ):
+      # These cases are not supported by get_text_unmarked
+      return
+
+    text_unmarked = self.atok.get_text_unmarked(node)
+    if isinstance(node, ast.alias):
+      self._check_alias_unmarked(node, test_case, text_unmarked)
+    elif not isinstance(node, self.bad_unmarked_types):
+      if hasattr(node, 'lineno'):
```
dsagal, line 135: How often is 'lineno' absent? (I guess I am worried if it might fail to compare a large set of cases, giving false confidence that `get_text_unmarked` works well.)

dsagal (approved):
Just one question about the test. Also, since f-strings are motivators, might it make sense to add a test case that specifically checks that `get_text_unmarked` works with f-strings?

tests/tools.py
alexmojaki: This is a good point. It's generally for 'unusual' nodes, but apparently that line is blurrier than I thought.

alexmojaki:
At this stage I mostly wanted to get feedback on the general plan and proposed API. Quite a bit more code needs to be written to fulfill what I've written above. Are you happy with that plan? Since you've accepted this PR, does that mean I should merge this and then write the rest in another PR?

dsagal (approved):
I am fine with the plan, and with landing this as is and leaving follow-ups for future PRs.

One question about bigger picture: for f-strings, does tokenizing not work at all? In a simple `generate_tokens` call, it doesn't seem to. Does that mean that code supporting f-strings should generally avoid any reliance on `start_token`/`end_token`? If so, the plan makes a lot of sense. If tokens do still exist in some sense, maybe it's possible to preserve the tokenizing behavior of `asttokens`, in which case the alternative approach with `get_text_unmarked()` seems a less useful substitute.

alexmojaki:
The primary motivation for me is to avoid the performance cost of tokenizing an entire file and then matching those tokens with nodes. I don't generally use information relating to tokens, I usually only care about AST nodes and their locations. Solving the problem of f-strings would be a nice bonus.

If people want to work with tokens within f-strings, then I think that requires obtaining the source code of the `ast.FormattedValue` and tokenizing that source in isolation. An easy way to get that source code is with `get_text_unmarked`. Otherwise you need some workaround like https://github.com/gristlabs/asttokens/issues/6#issuecomment-504760714.

Once you have that inner set of tokens, you still have to deal with the fact that they're somewhat separate from the original outer tokens. Do you merge everything into one big set? Do you discard the 'official' Python token corresponding to the entire f-string? Do you invent new tokens for the curly brackets and the intervening constant parts of the string? Do you adjust each `Token.index`? Or if you keep the sets of tokens separate, do you define new semantics for `next_token` and `prev_token`?

PR #93, opened 2022-09-10
ASTText class that doesn't require tokens

### LINKED ISSUE #97: Consider a clearer optional-functionality API
Currently `ASTTokens` serves a number of different roles depending on what arguments its passed. With https://github.com/gristlabs/asttokens/pull/93 this is growing further and the workarounds to type checking being added in that PR perhaps suggest that the limits of the current approach are being reached. https://github.com/gristlabs/asttokens/pull/95 will add another construction, though not another role.
I understand the goals of these different roles are around avoiding redundant parsing/tree-building work where not needed, which is a reasonable goal. I'm not suggesting a removal of functionality here, rather that the functionality be rearranged for clarity and to aid type checking.

There are at least the following modes of use:
``` python
atok = ASTTokens(source, parse=True)
assert atok.tree, "Placate type check even though we know this is always valid"

atok = ASTTokens(source, tree=existing_tree)
assert atok.tree, "Placate type check even though we know this is always valid"
```

There is also currently `ASTTokens(source, tree=None, parse=False)`, though it's not clear what use that has (https://github.com/gristlabs/asttokens/issues/94#issuecomment-1252862533).

With #93 there's also this case:
``` python
atok = ASTTokens(source, init_tokens=False)
atok.get_token_from_offset(42)  # Fail at runtime, but no type-check warnings
```

These cases are somewhat "obvious", however if the construction of the `ASTTokens` instance is far from the point of use they could easily be very non-obvious.

We can observe that `ASTTokens` contains broadly three classes of functionality:
- `tokens`-only
- `tree`-only
- both together

With the goal of enabling efficient use of the related functionalities I can see a couple of paths forwards:
- lazy evaluation of the relevant structured data from the source
- failures when using functionality with missing data

For the former I would advocate for moving all the public-member types to being non-optional and using something like https://pypi.org/project/cached-property/ (or [`functools.cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) in Python 3.8+) to hide the evaluations. This would still support users passing in pre-computed values if they want but means that all usages will "just work":
``` python
atok = ASTTokens(source)
tree: ast.AST = atok.tree  # no error now that `.tree` is non-optional

atok = ASTTokens(source, init_tokens=False)  # request lazy tokens initialisation
atok.get_token_from_offset(42)  # works due to lazy parsing
```

I can see that this approach may be undesirable in some cases, perhaps where there are reasons to prefer not doing the parsing _at all_.

This leads to the other approach (errors for missing functionality), for which I propose using separate classes:

``` python
# ASTSource is a thin wrapper around LineNumbers. (Very open to better names.)
# Does not take a tree or tokens at all, but works with un-marked AST nodes
asrc = ASTSource(source)
asrc.tree  # type check error: no such member
asrc.get_text(node)  # type check error: no such member
asrc.get_token_from_offset(42)  # type check error: no such member
asrc.get_text_unmarked(node)  # works fine

# TokensSource is a new type which contains the tokens/source utils currently in ASTTokens
# It wraps LineNumbers similarly to ASTSource, but is otherwise unrelated
# It does not have a tree
tsrc = TokensSource(source)  # parses tokens if not passed
tsrc.tree  # type check error: no such member
tsrc.get_text(node)  # type check error: no such member
tsrc.get_token_from_offset(42)  # works fine
tsrc.get_text_unmarked(node)  # type check error: no such member

# ASTTokens now always has a tree and tokens, creating them both if not required
# It probably inherits from TokensSource and drops the `get_text*_unmarked` methods as they no longer make sense here
atok = ASTTokens(source)
tree: ast.AST = atok.tree  # no error now that `.tree` is non-optional
atok.get_text(node)  # works fine
atok.get_token_from_offset(42)  # also works fine
atok.get_text_unmarked(node)  # type check error: no such member
```

Aside from the question of whether `ASTTokens` keeps its `get_text*_unmarked` methods, the usages here are essentially compatible with existing code. The constructor signature might change, though keeping the existing arguments for compatibility is probably possible; if so then `ASTTokens(source, tree=None, parse=False)` would now be at least a runtime error (with `@overload`s it could be a type error too).

Note also that this approach is compatible with another class, perhaps `LazyASTTokens` which looks like `ASTTokens` but has lazy evaluation. This could easily be added later (or not at all) depending on whether it ends up being useful.

The main gain here is that each of these specialised classes addresses a specific use-case and data-requirements, which remove the need for conditional presence checks (or asserts) throughout both the library code and client code. This will make the code clearer, easier to maintain and (due to the type checker support) more robust as well as offering the performance gains desired.

alexmojaki:
Following up on https://github.com/gristlabs/asttokens/pull/92.

Closes #97.

- Adds a new class `ASTText` that:
  - Allows using `get_text[_range]` while avoiding the performance cost of dealing with tokens for most node types and newer Python versions.
  - Supports nodes inside f-strings for newer Python versions.
  - Lazily initialises `tree` if one isn't passed, so `parse=True` isn't needed.
- Adds a new method `get_text_positions` to both `ASTText` and `ASTTokens` which returns start and end (lineno, col_offset) pairs, as an alternative to `node.first/last_token`.
- Adds a boolean argument `padded` to all 3 `get_text*` methods on both classes which matches the behaviour of the same argument in `ast.get_source_segment` i.e. includes leading indentation for multiline statements. However the default is True (instead of False as in `ast.get_source_segment`) in the existing methods `get_text[_range]` for backwards compatibility. For the new `get_text_positions` there is no default to force the user to stop and think. It can't be `False` because that would be confusing compared to the existing methods, but `False` matches the behaviour of using `node.first/last_token` which the method is meant to replace.

### DIFF
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -236,7 +236,11 @@ class ASTTokens(object):
     """
     TODO
     """
-    if unmarked or (not self._tokens and supports_unmarked(node)):
+    in_f_string = getattr(node, "_in_f_string", False)
+    if unmarked or (
+        (not self._tokens or in_f_string)
+        and supports_unmarked(node)
+    ):
       return self._get_text_positions_unmarked(node, padded)
 
     self.init_tokens()
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -25,7 +25,7 @@ import six
 from six.moves import xrange  # pylint: disable=redefined-builtin
 
 from .line_numbers import LineNumbers
-from .util import Token, match_token, is_non_coding_token, patched_generate_tokens, last_stmt
+from .util import Token, match_token, is_non_coding_token, patched_generate_tokens, last_stmt, annotate_fstring_nodes
 
 if TYPE_CHECKING:
   from .util import AstNode
@@ -67,6 +67,8 @@ class ASTTokens(object):
 
     self._filename = filename
     self._tree = ast.parse(source_text, filename) if parse else tree
+    if self._tree is not None:
+      annotate_fstring_nodes(self._tree)
 
     # Decode source after parsing to let Python 2 handle coding declarations.
     # (If the encoding was not utf-8 compatible, then even if it parses correctly,
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -33,7 +33,7 @@ if TYPE_CHECKING:  # pragma: no cover
 class ASTTextBase(six.with_metaclass(abc.ABCMeta, object)):
   def __init__(self, source_text, filename):
     # type: (Any, str) -> None
-    # FIXME: Strictly, the type of source_type is one of the six string types, but hard to specify with mypy given
+    # FIXME: Strictly, the type of source_text is one of the six string types, but hard to specify with mypy given
     # https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
 
     self._filename = filename
@@ -99,7 +99,7 @@ class ASTTokens(ASTTextBase, object):
 
   def __init__(self, source_text, parse=False, tree=None, filename='<unknown>', tokens=None):
     # type: (Any, bool, Optional[Module], str, Iterable[TokenInfo]) -> None
-    # FIXME: Strictly, the type of source_type is one of the six string types, but hard to specify with mypy given
+    # FIXME: Strictly, the type of source_text is one of the six string types, but hard to specify with mypy given
     # https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
 
     super(ASTTokens, self).__init__(source_text, filename)
@@ -280,7 +280,7 @@ class ASTTokens(ASTTextBase, object):
 class ASTText(ASTTextBase, object):
   def __init__(self, source_text, tree=None, filename='<unknown>'):
     # type: (Any, Optional[Module], str) -> None
-    # FIXME: Strictly, the type of source_type is one of the six string types, but hard to specify with mypy given
+    # FIXME: Strictly, the type of source_text is one of the six string types, but hard to specify with mypy given
     # https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
 
     if not isinstance(tree, (ast.AST, type(None))):
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -260,6 +260,11 @@ class ASTTokens(object):
     if not supports_unmarked():
       raise NotImplementedError('Python version not supported')
 
+    # This is already checked above, but this check is recognized by mypy.
+    # This prevents errors below with end_lineno and end_col_offset.
+    if TYPE_CHECKING and sys.version_info[:2] < (3, 8):
+      raise NotImplementedError('Python version not supported')
+
     if not isinstance(node, ast.AST):
       raise NotImplementedError('Not supported for astroid')
 
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -172,7 +172,6 @@ class ASTTokens(ASTTextBase, object):
     Returns the token containing the given character offset (0-based position in source text),
     or the preceeding token if the position is between tokens.
     """
-    assert self._token_offsets and self._tokens
     return self._tokens[bisect.bisect(self._token_offsets, offset) - 1]
 
   def get_token(self, lineno, col_offset):
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -249,7 +249,8 @@ class ASTTokens(object):
     # type: (AstNode, bool, bool) -> Tuple[Tuple[int, int], Tuple[int, int]]
     """
     Returns two ``(lineno, col_offset)`` tuples for the start and end of the given node.
-    If the positions can't be determined, returns ``(1, 0)`` for both.
+    If the positions can't be determined, or the nodes don't correspond to any particular text,
+    returns ``(1, 0)`` for both.
 
     ``padded`` corresponds to the ``padded`` argument to ``astunparse.get_source_segment()``.
     This means that if ``padded`` is True, the start position will be adjusted to include
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -238,7 +238,7 @@ class ASTTokens(object):
     """
     in_f_string = getattr(node, "_in_f_string", False)
     if unmarked or (
-        (not self._tokens or in_f_string)
+        (not self._tokens or in_f_string or isinstance(node, (ast.alias)))
         and supports_unmarked(node)
     ):
       return self._get_text_positions_unmarked(node, padded)
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -231,13 +231,13 @@ class ASTTokens(object):
     """
     return self.token_range(node.first_token, node.last_token, include_extra=include_extra)
 
-  def get_text_positions(self, node, unmarked=False):
-    # type: (AstNode, bool) -> Tuple[Tuple[int, int], Tuple[int, int]]
+  def get_text_positions(self, node, padded, unmarked=False):
+    # type: (AstNode, bool, bool) -> Tuple[Tuple[int, int], Tuple[int, int]]
     """
     TODO
     """
     if unmarked or (not self._tokens and supports_unmarked(node)):
-      return self._get_text_positions_unmarked(node)
+      return self._get_text_positions_unmarked(node, padded)
 
     self.init_tokens()
 
@@ -246,14 +246,13 @@ class ASTTokens(object):
 
     start = node.first_token.start
     end = node.last_token.end
-    if any(match_token(t, token.NEWLINE) for t in self.get_tokens(node)):
-      # Multi-line nodes would be invalid unless we keep the indentation of the first node.
+    if padded and any(match_token(t, token.NEWLINE) for t in self.get_tokens(node)):
       start = (start[0], 0)
 
     return start, end
 
-  def _get_text_positions_unmarked(self, node):
-    # type: (AstNode) -> Tuple[Tuple[int, int], Tuple[int, int]]
+  def _get_text_positions_unmarked(self, node, padded):
+    # type: (AstNode, bool) -> Tuple[Tuple[int, int], Tuple[int, int]]
     """
     TODO
     """
@@ -284,9 +283,7 @@ class ASTTokens(object):
     else:
       start_node = node
 
-    # Like get_text_range(), keep the indentation of the first line
-    # of a multi-line, multi-statement node.
-    if last_stmt(node).lineno != node.lineno:
+    if padded and last_stmt(node).lineno != node.lineno:
       start_col_offset = 0
     else:
       start_col_offset = self._line_numbers.from_utf8_col(start_node.lineno, start_node.col_offset)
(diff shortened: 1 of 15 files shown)


PeterJCLaw:
Just a thought here -- would using something like https://pypi.org/project/cached-property/ make this easier given the goal of lazy evaluation? That would sidestep the need for various places to check whether the tokens have been initialised (I'm imagining that consumers would just access `self.tokens` or `self.token_offsets` which would facade out the lazy-parsing side of things) and make the type checkers happier too.

Not sure how big an issue it would be to take the extra dependency for Python < 3.8 though? (`functools` learns [`cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) in 3.8)

If that approach does work, given the conversation at https://github.com/gristlabs/asttokens/issues/94#issuecomment-1252862533 would it make sense to make the AST similarly lazy (rather than potentially `None`)?

alexmojaki:
The `tokens` property isn't lazily initialising `_tokens`, it's asserting it for the sake of type checkers. Using `.tokens` when they're not initialised is meant to throw an error.

PeterJCLaw:
Ah, ok. I think I'd misunderstood from the discussion on that issue.

Given that goal I would repeat the encouragement I made on that issue -- of exploring having variations of `ASTTokens` for the no-tokens and no-tree cases so that the default case requires both. This would greatly simplify the type checking and in turn make consumer code clearer. I'd be happy to put together a sketch of that if you're open to that approach.

alexmojaki:
Can you show a sketch of what the consumer code would look like and what the difference/advantage would be? Right now I think the default case is `atok = ASTTokens(source, parse=True)` and I don't see much of a problem that needs improving.

PeterJCLaw:
> Can you show a sketch of what the consumer code would look like and what the difference/advantage would be? Right now I think the default case is `atok = ASTTokens(source, parse=True)` and I don't see much of a problem that needs improving.

Sure, though I'll take it to a separate issues to avoid taking this PR too far off topic.

PeterJCLaw:
> > Can you show a sketch of what the consumer code would look like and what the difference/advantage would be? Right now I think the default case is `atok = ASTTokens(source, parse=True)` and I don't see much of a problem that needs improving.
> 
> Sure, though I'll take it to a separate issues to avoid taking this PR too far off topic.

https://github.com/gristlabs/asttokens/issues/97

alexmojaki:
@dsagal the names I chose initially are not great. What do you think of replacing `unmarked` and `init_tokens` with `use_tokens`? So the constructor and get_text methods would both have `use_tokens=True` by default, and the method `.init_tokens()` would be `.use_tokens()`. `supports_unmarked` could be `supports_no_tokens`.

alexmojaki:
Here's a comparison to just before phase 1 (#92) to show the whole picture which can be helpful: https://github.com/gristlabs/asttokens/compare/2.0.8...unmarked2

alexmojaki:
Found a problem while testing this in futurecoder. I had some code that relied on the fact that `.get_text(node)` returning something non-empty means that `node.first_token` exists. The new way of supporting f-strings means that this implication no longer holds, even without setting `init_tokens=False`. This led to an AttributeError. I don't see anything we can do about that here, so that's a small break in compatibility that probably means a new major version is needed.

alexmojaki:
Two more f-string problems:

1. `.get_text(node)` suddenly working can just break things in other ways. Previously I was excluding nodes from a particular set when `get_text` didn't work. f-string contents suddenly appearing in the set was a problem. In particular, I was using `asttokens.util.replace` to modify some source code, and it assumes non-overlapping ranges, but now the inner node ranges were overlapping with the full f-string range. The fact that I was using an `asttokens.util` function here is essentially a coincidence, things could just break in non-asttokens related ways. In theory I suppose there could be another boolean arg for compatibility to continue ignoring f-string nodes, but that seems a bit extreme.
2. The direct children of f-strings (but not the regular expression nodes within) have broken positions and need further special handling.

asttokens/asttokens.py
```diff
@@ -213,86 +245,160 @@ def get_tokens(self, node, include_extra=False):
...
+    if getattr(node, "_broken_positions", None):
+      return (1, 0), (1, 0)
+
+    node_dont_use_tokens = getattr(node, "_dont_use_tokens", None)
+    # Always use the unmarked method when unmarked=True.
+    if unmarked or (
+        # For nodes that work with the unmarked method, prefer it if either:
+        # 1. There are no tokens, i.e. `init_tokens=False` was passed to the constructor.
+        # 2. The node is better handled by the unmarked method. This applies to import aliases
+        #    and nodes within f-strings.
```
PeterJCLaw, line 272: Looks like this applies to more than just f-strings now?

alexmojaki: It seemed best to make `_dont_use_tokens` a generic name that could theoretically be applied to other things, but for now it only applies to nodes within f-strings.

PeterJCLaw:
While rebasing #98 I hit a test failure which I think is present in this PR too -- the f-string handling in `atok.get_text` implicitly relies on `mark_tokens` having run before it's called. This seems to happen in the example source because of the function definition there, however commenting that out (the `def foo` on line 16--17 in `tests/test_unmarked.py`) causes `test_unmarked` to fail on the f-string handling.

The specific error comes from the assertion that `atok.get_text` for f-strings return empty text: `self.assertEqual(atok_text, "", node)`.

alexmojaki:
I've explained the reason for the conditional block in https://github.com/gristlabs/asttokens/pull/98#discussion_r985298862. However you're right about this:

>  the f-string handling in atok.get_text implicitly relies on mark_tokens having run before it's called

Initially this was intentional and fine because the only f-string handling was `_in_f_string` (now called `_dont_use_tokens`) saying that even though we have tokens you still shouldn't use them.

However it was a mistake to also add `_broken_positions` in the same place, because indeed this means that marking tokens changes the behaviour of `get_text`. The tests actually do reflect this now but I forgot about this point.

So another solution is needed if we want the 'broken' inner nodes of f-strings to return empty text instead of the whole f-string. There's 3 types of these nodes:

- FormattedValue: easy, this type seems to always be broken.
- Str/Constant: quite easy if we just notice that the returned text starts with `f` or `rf` (or `F` etc.) instead of `"` or `'`.
- JoinedStr: this is for the format spec. This is a problem, since the returned text is the full f-string which is the correct type. So I think we either have to:
  - Iterate through the whole tree on initialisation to mark nodes. Not great since until now we managed to avoid paying any similar upfront cost.
  - Parse the returned text and see if it matches the node we have. Makes each `get_text` call potentially significantly more expensive, but only for JoinedStr nodes.

dsagal:
> @dsagal the names I chose initially are not great. What do you think of replacing `unmarked` and `init_tokens` with `use_tokens`? So the constructor and get_text methods would both have `use_tokens=True` by default, and the method `.init_tokens()` would be `.use_tokens()`. `supports_unmarked` could be `supports_no_tokens`.

Sure! I haven't followed the full discussion on this or the related issues. I'd be fine with different names for params/methods. Having separate classes for different semantics seems reasonable too, especially if that helps with typings.

alexmojaki:
@dsagal aside from the broader API questions, it'd be helpful if you could look at my last comment above about f-strings, including the comment I linked to showing the general problem, in case you have any good ideas for solutions, or a reason why it's not worth worrying about.

dsagal:
> @dsagal aside from the broader API questions, it'd be helpful if you could look at my last comment above about f-strings, including the comment I linked to showing the general problem, in case you have any good ideas for solutions, or a reason why it's not worth worrying about.

It seems to me like f-strings are an area similar to where the AST itself was when `asttokens` was created: the built-in functionality has bugs that make it insufficient, and to work around them, `asttokens` needs to do some manual parsing (searching for curly braces, `:`, `!`). Sorry, I don't have anything more helpful to suggest...

alexmojaki:
Well since you noticed that arbitrary types of expression nodes inside f-strings can have wrong positions, I think that settles the fact that walking through the tree to mark various nodes is necessary.

PeterJCLaw:
> @PeterJCLaw please can you write an example of what using the API in #98 would look like, compared with the API here,

I think the tests in #98 cover this? Essentially construction is similar to that of plain `ASTTokens`, however usage of the lazy class is limited to the functionality that doesn't "obviously" need tokens.

I'm not really expecting consumers to use the `LazyHelper.asttokens` property, though they could if they wanted (and would get what they expected in return).

> and a summary of the benefits?

The main benefits as I see them are:
- `ASTTokens` and `LazyHelper` have useful static analysis (which the consequences of `ASTTokens(init_tokens=False)` does not), allowing both type checkers and editor completions to be essentially correct without additional effort (including any changes necessary due to opting in to the functionality in this PR #93)
- The semantics of what is possible with each class are consequently clearer as they're static (indeed this was how the f-strings bug was highlighted), resulting in clearer behaviour for both consumers and maintainers of this package (and in turn hopefully easier maintenance)
- The semantics of _all_uses of `LazyHelper` are much more obvious in terms of their impact on `node.{first,last}_token` -- specifically that consumers should not use those added attributes on the grounds that they might not be present (but that's what consumers are opting in to by using this class)
- (`LazyHelper` also happens to allow for the case of lazily evaluating the AST too, meaning other use-cases can choose to benefit from laziness without this functionality further complicating the `ASTTokens` initialiser)
- Consumers that really want to access the additional `ASTTokens` functionality can still do so (and the point at which the lazy-loading happens will be obvious to them as a result of their needing to access an indirect property to do so)
- Easier maintenance as logically separate code is in separate places (including e.g: the `TextRangeBase` class) and in particular the optimising/special-casing being added for the lazy case are limited to one specific place in the code (the `LazyHelper` class)

The potential drawbacks I can see are:
- To opt in to the new behaviour consumers need to use a new class:
  - potentially needing more changes than just the initialiser argument (minor IMO given that they'd be changing code anyway)
  - counterpoint: most additional changes they need to make will be highlighted to them by a type checker (if they're using one)
- Some mental overhead during maintenance from having the logic split between multiple classes

alexmojaki:
With @PeterJCLaw's changes:

- The `ASTTokens` class only uses tokens and doesn't use the new method at all.
- This means that it can't handle f-strings at all again.
- The previously mentioned compatibility problems are gone.
- The special attribute `_dont_use_tokens` (from earlier in this PR) is no longer needed and has been removed.

asttokens/asttokens.py
```diff
@@ -209,86 +258,159 @@ def get_tokens(self, node, include_extra=False):
...
+  if sys.version_info[:2] == (3, 8):
+    _unsupported_no_tokens_types += (
+      # _get_text_positions_no_tokens works incorrectly for these types due to bugs in Python 3.8.
+      ast.arg, ast.Starred,
+      # no lineno in 3.8
+      ast.Slice, ast.ExtSlice, ast.Index, ast.keyword,
+    )
+
+
+def supports_no_tokens(node=None):
```
dsagal, line 404: For the name, it sounds like either "supports operation without tokens" or "doesn't support tokens". Would the name "supports_tokenless" be clearer?

asttokens/asttokens.py
```diff
@@ -209,86 +258,159 @@ def get_tokens(self, node, include_extra=False):
...
+  # type: (Any) -> bool
+  """
+  Returns True if the Python version and the node (if given) are supported by
+  the ``get_text*`` methods of ``ASTText`` without falling back to token
```
dsagal, line 408: Is the point performance? Or better behavior? Either way, an explanation of purpose would be helpful.

dsagal (commented):
I haven't reviewed in full, but including initial comments. Since I am catching up on the changes, it would help to have a clearer name and description of the goals of this PR.

dsagal (commented):
All makes sense, thanks! I have a few minor notes, mostly about adding comments to code.

dsagal (approved):
Thanks!

Issue #94, opened 2022-09-19
Support for constructing around a list of existing tokens


PeterJCLaw:
Thanks for this library, I've found it really useful for building source manipulation tools.

I'm currently building a flake8 plugin and it would be great if there was a way to build an `ASTTokens` instance from an existing list of (stdlib) tokens. This would complement the ability to pass in an existing AST tree and allow cases which already have both available to avoid re-parsing the source code. (Flake8 makes both an existing AST and tokens list available to plugins).

If this is something you'd be open to I'd be happy to put together a PR. (The change seems simple enough, though possibly I'm missing something)

alexmojaki:
Sounds sensible. You should probably have a brief look at the first few changes to `asttokens.py` in #93 as they will be relevant to this. I imagine the ultimate solution will simply be converting the variable `original_tokens` in `_generate_tokens` to an optional parameter of `_generate_tokens`, `init_tokens`, and `__init__`.

PeterJCLaw:
> I imagine the ultimate solution will simply be converting the variable `original_tokens` in `_generate_tokens` to an optional parameter of `_generate_tokens`, `init_tokens`, and `__init__`.

Yup, that's pretty much what I had in mind.

> You should probably have a brief look at the first few changes to `asttokens.py` in #93 as they will be relevant to this.

Ah, yeah, that's going to overlap a bit and I suspect that allowing all the various construction cases is going to be a bit complicated too.

For my understanding: what use is the `ASTTokens` class if it doesn't have an AST to work with (i.e: both `tree=None` and `parse=False`)? Are there consumers who use it just for the token utils? (I've found them useful, though wouldn't have thought to use this library just for them).

I realise this is perhaps a bit beyond the scope of this issue and I'm perhaps late to the conversation (and a breaking API change), but I wonder whether it would make sense to have `ASTTokens` always build up the aspects (tokens/ast) that it doesn't have rather than trying to accommodate the cases of partial data?

That would allow the `ASTTokens` class to be simpler -- a lot of the `assert`s that PR adds would go away and consumers wouldn't need to type-guard access to `ASTTokens.tree`.

If the token-only or ast-only functionality is useful on its own then perhaps extracting them to standalone functions or base classes would make sense? That could leave the current `ASTTokens` class pulling all the features together for those that want it, but allow subsets to be used on their own without complicating the code.

If you'd rather stick with the current design I'm happy to work with that, I just didn't want to end up creating a lot of complications for #93.

[...]

PR #95, opened 2022-09-20
Support passing in tokens already parsed from the source

PeterJCLaw:
Fixes https://github.com/gristlabs/asttokens/issues/94

### DIFF
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -25,7 +25,7 @@ from six.moves import xrange  # pylint: disable=redefined-builtin
 from .line_numbers import LineNumbers
 from .util import Token, match_token, is_non_coding_token, patched_generate_tokens, last_stmt, generate_tokens
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
   from .util import AstNode, TokenInfo
 
 
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -19,7 +19,7 @@ import sys
 import token
 import tokenize
 from ast import Module
-from typing import Callable, Iterator, List, Optional, Tuple, Any, cast, TYPE_CHECKING
+from typing import Callable, Iterable, Iterator, List, Optional, Tuple, Any, cast, TYPE_CHECKING
 
 import six
 from six.moves import xrange  # pylint: disable=redefined-builtin
@@ -28,7 +28,7 @@ from .line_numbers import LineNumbers
 from .util import Token, match_token, is_non_coding_token, patched_generate_tokens, last_stmt
 
 if TYPE_CHECKING:
-  from .util import AstNode
+  from .util import AstNode, TokenInfo
 
 
 class ASTTokens(object):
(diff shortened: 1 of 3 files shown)


asttokens/util.py
```diff
@@ -320,11 +325,11 @@ def get(self, obj, cls):
...
   # means working with raw tuples instead of tokenize.TokenInfo namedtuples.
   def patched_generate_tokens(original_tokens):
-    # type: (Any) -> Any
-    return original_tokens
+    # type: (Iterable[TokenInfo]) -> Iterator[TokenInfo]
+    return iter(original_tokens)
 else:
   def patched_generate_tokens(original_tokens):
-    # type: (Iterator[tokenize.TokenInfo]) -> Iterator[tokenize.TokenInfo]
+    # type: (Iterable[tokenize.TokenInfo]) -> Iterator[tokenize.TokenInfo]
```
alexmojaki, line 332:
```suggestion
    # type: (Iterable[TokenInfo]) -> Iterator[TokenInfo]
```

just for consistency

tests/test_asttokens.py
```diff
@@ -34,6 +36,38 @@ def test_tokenizing(self):
     self.assertEqual(atok.tokens[5].endpos, 25)
 
 
+  def test_given_existing_tokens(self):
```
alexmojaki, line 39: This is almost identical to the previous test, which makes sense. Please deduplicate the code.

It would also be nice to have a test of what happens when you pass tokens that aren't identical to what's produced normally.

alexmojaki (changes requested):
General approach looks good!

tests/test_asttokens.py
PeterJCLaw: > It would also be nice to have a test of what happens when you pass tokens that aren't identical to what's produced normally.

I'm not sure I follow; do you have an example which would clarify this?
Do you mean for example passing incorrect tokens which don't match the source? Or something else?

alexmojaki: Incorrect tokens would do the trick, I can't think of any other way. I partly just want something that shows that it really used the passed in tokens rather than generating its own.

asttokens/asttokens.py
```diff
@@ -90,14 +92,21 @@ def mark_tokens(self, root_node):
 
 
   def _generate_tokens(self, text):
```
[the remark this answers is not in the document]
PeterJCLaw: https://github.com/gristlabs/asttokens/pull/95/commits/f927445cc1d8b3ef8e8f175acfba764881a99ef2 does this.

tests/test_asttokens.py
PeterJCLaw: https://github.com/gristlabs/asttokens/pull/95/commits/363c1323ba6183dd9b7060f63edb22986bf59575 deduplicates, https://github.com/gristlabs/asttokens/pull/95/commits/fa69685b81841f3b249571352b70bc8241ad9b0f validates the tokens are used.

asttokens/util.py
```diff
@@ -36,6 +37,11 @@ class EnhancedAST(AST):
 
   AstNode = Union[EnhancedAST, NodeNG]
 
+  if sys.version_info[0] == 2:
```
alexmojaki, line 40: Please add `  # pragma: no cover` to `if TYPE_CHECKING:` above to solve the reduction in coverage in this PR.

PeterJCLaw: 4e931c6bb2b9ef683761a544c1645a8d12fd9b67

alexmojaki:
Can you convert this into another PR on #93?

PeterJCLaw:
> Can you convert this into another PR on #93?

I can if that'll help avoid conflicts, though it feels like it should probably be merged separately given its different goals?

alexmojaki:
You're right, it wouldn't really help. Can you resolve the conflicts please?

alexmojaki:
Also @dsagal please can you give @PeterJCLaw collaborator access? I think he's earned it.

dsagal:
Just did. Welcome, @PeterJCLaw, and thank you for contributing!!

PR #98, opened 2022-10-02
Explore a clearer lazy-parsing behaviour API

PeterJCLaw:
This introduces a new class which has the lazy-parsing behaviour, separate from `ASTTokens`, thus leaving `ASTTokens` semantics unchanged.

`ASTTokens` still learns how to handle nodes unmarked, however it is left somewhat cautious in the use of them. The new class instead prefers them wherever it can.

This aims to provide an alternative to the API from #93 as the current API would make static analysis of `ASTTokens` much less useful when the `init_tokens` argument is used.

While it looks like there's a lot of change here, there isn't -- most of the diff is moving functions to the top of the file.

### DIFF
--- a/asttokens/asttokens.py
+++ b/asttokens/asttokens.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import abc
 import ast
 import bisect
 import io
@@ -31,44 +32,13 @@ if TYPE_CHECKING:
   from .util import AstNode
 
 
-class ASTTokens(object):
-  """
-  ASTTokens maintains the text of Python code in several forms: as a string, as line numbers, and
-  as tokens, and is used to mark and access token and position information.
-
-  ``source_text`` must be a unicode or UTF8-encoded string. If you pass in UTF8 bytes, remember
-  that all offsets you'll get are to the unicode text, which is available as the ``.text``
-  property.
-
-  If ``parse`` is set, the ``source_text`` will be parsed with ``ast.parse()``, and the resulting
-  tree marked with token info and made available as the ``.tree`` property.
-
-  If ``tree`` is given, it will be marked and made available as the ``.tree`` property. In
-  addition to the trees produced by the ``ast`` module, ASTTokens will also mark trees produced
-  using ``astroid`` library <https://www.astroid.org>.
-
-  If only ``source_text`` is given, you may use ``.mark_tokens(tree)`` to mark the nodes of an AST
-  tree created separately.
-
-  If ``init_tokens`` is False, the text will not be tokenized and the nodes won't be marked.
-  This can save significant CPU time. In this case token-specific methods such as ``.get_token``
-  will raise an error, but position-related methods such as ``.get_text(node)`` will still work.
-  You can still call ``.init_tokens()`` later to tokenize the text and mark the nodes.
-  ``init_tokens=False`` will be ignored for Python < 3.8 or PyPy.
-  It will raise an error if ``tree`` is an astroid tree.
-  Finally, note that methods such as ``.get_text(node)`` may still automatically call ``.init_tokens()``
-  for specific unusual types of node that aren't supported.
-  """
-
-  def __init__(self, source_text, parse=False, tree=None, filename='<unknown>', init_tokens=True):
-    # type: (Any, bool, Optional[Module], str, bool) -> None
+class ASTTextBase(six.with_metaclass(abc.ABCMeta, object)):
+  def __init__(self, source_text, filename):
+    # type: (Any, str) -> None
     # FIXME: Strictly, the type of source_type is one of the six string types, but hard to specify with mypy given
     # https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
 
     self._filename = filename
-    self._tree = ast.parse(source_text, filename) if parse else tree
-    if self._tree is not None:
-      annotate_fstring_nodes(self._tree)
 
     # Decode source after parsing to let Python 2 handle coding declarations.
     # (If the encoding was not utf-8 compatible, then even if it parses correctly,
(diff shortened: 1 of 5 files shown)


asttokens/asttokens.py
```diff
@@ -31,7 +32,111 @@
...
+
+    self._text = source_text
+    self._line_numbers = LineNumbers(source_text)
+
+  @abc.abstractmethod
+  def get_text_positions(self, node, padded, unmarked=False):
+    # type: (AstNode, bool, bool) -> Tuple[Tuple[int, int], Tuple[int, int]]
+    raise NotImplementedError
+
+  def _get_text_positions_unmarked(self, node, padded):
```
PeterJCLaw, line 56: This and the following methods on this class are moved verbatim from `ASTTokens`.

asttokens/asttokens.py
```diff
@@ -285,87 +349,58 @@ def get_text_positions(self, node, padded, unmarked=False):
...
-    if isinstance(node, ast.Module):
-      # Modules don't have position info, so just return the range of the whole text.
-      # The token-using method does something different, but its behavior seems weird and inconsistent.
-      # For example, in a file with only comments, it only returns the first line.
-      # It's hard to imagine a case when this matters.
-      return (1, 0), self._line_numbers.offset_to_line(len(self._text))
-
-    if not hasattr(node, 'lineno'):
-      return (1, 0), (1, 0)
+class LazyHelper(TextRangeBase, object):
```
PeterJCLaw, line 353: Very much open to better names; consider this a placeholder.

asttokens/asttokens.py
```diff
@@ -285,87 +349,58 @@ def get_text_positions(self, node, padded, unmarked=False):
...
+  def __init__(self, source_text, tree=None, filename='<unknown>'):
+    # type: (Any, Optional[Module], str) -> None
+    # FIXME: Strictly, the type of source_type is one of the six string types, but hard to specify with mypy given
+    # https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
 
-    assert node  # tell mypy that node is not None, which we allowed up to here for compatibility
+    if not isinstance(tree, (ast.AST, type(None))):
+      raise NotImplementedError('LazyHelper only supports AST trees')
```
PeterJCLaw, line 360: I suspect with tweaks to how the `unmarked` parameters are handled this limitation could be dropped. Having it keeps the tests essentially unchanged and it'd be easier to remove the limitation later than to add it back in.

alexmojaki:
Thanks, this looks promising. My PR is stable again, you can resolve the conflicts.

tests/test_unmarked.py
```diff
@@ -75,41 +75,34 @@ def test_unmarked(self):
...
       ast_text = ast.get_source_segment(source, node, padded=True)
       if ast_text:
-        if (
-            ast_text.startswith("f") and isinstance(node, (ast.Str, ast.FormattedValue))
-            or is_fstring_format_spec(node)
-        ):
-          self.assertEqual(atok_text, "", node)
-        else:
-          self.assertEqual(atok_text, ast_text, node)
+        self.assertEqual(atok_text, ast_text, node)
```
alexmojaki, line 85: The issue is that `ast_text` is a mess in the first place. Here's a demo:

```python
import ast

source = 'f"a{b:{c:d}}"'

tree = ast.parse(source)
for node in ast.walk(tree):
  if ast.get_source_segment(source, node) == source:
    print(ast.dump(node))
```

Output:

```python
Expr(value=JoinedStr(values=[Constant(value='a'), FormattedValue(value=Name(id='b', ctx=Load()), conversion=-1, format_spec=JoinedStr(values=[FormattedValue(value=Name(id='c', ctx=Load()), conversion=-1, format_spec=JoinedStr(values=[Constant(value='d')]))]))]))
JoinedStr(values=[Constant(value='a'), FormattedValue(value=Name(id='b', ctx=Load()), conversion=-1, format_spec=JoinedStr(values=[FormattedValue(value=Name(id='c', ctx=Load()), conversion=-1, format_spec=JoinedStr(values=[Constant(value='d')]))]))])
Constant(value='a')
FormattedValue(value=Name(id='b', ctx=Load()), conversion=-1, format_spec=JoinedStr(values=[FormattedValue(value=Name(id='c', ctx=Load()), conversion=-1, format_spec=JoinedStr(values=[Constant(value='d')]))]))
JoinedStr(values=[FormattedValue(value=Name(id='c', ctx=Load()), conversion=-1, format_spec=JoinedStr(values=[Constant(value='d')]))])
FormattedValue(value=Name(id='c', ctx=Load()), conversion=-1, format_spec=JoinedStr(values=[Constant(value='d')]))
JoinedStr(values=[Constant(value='d')])
Constant(value='d')
```

There's 6 AST nodes that all say that their source text spans the full f-string, when IMO none of them should. Does that seem OK to you?

PeterJCLaw: Ah, yeah, that's not great.
This feels like it would be worth some separate tests, rather than just bundled into the integration tests.

dsagal: Eek, yes, `get_source_segment` seems wrong, not only for the nodes for which it returns the full f-string, but also for nodes for which it doesn't. E.g. the first `Name` node, I imagine, should produce `'b'`, but is instead producing `'"'` (at least on Python 3.9.5)

alexmojaki: Oh, that's bad, well spotted. So yes, we'd need to do something about that too. I should also report this to Python if it hasn't been already. Here's a script:

```python
import ast

source = '''
f"a b { c + d : hgg } e f { g * 3 } h"
'''

tree = ast.parse(source)
for node in ast.walk(tree):
    segment = ast.get_source_segment(source, node)
    if not segment or segment.strip() == source.strip():
      continue
    print(ast.dump(node))
    print(segment)
    print()
```

Output:

```python
BinOp(left=Name(id='c', ctx=Load()), op=Add(), right=Name(id='d', ctx=Load()))
a b {

BinOp(left=Name(id='g', ctx=Load()), op=Mult(), right=Constant(value=3))
g * 3

Name(id='c', ctx=Load())
a

Name(id='d', ctx=Load())
{

Name(id='g', ctx=Load())
g

Constant(value=3)
3
```

Seems to be fixed in 3.10:

```python
BinOp(left=Name(id='c', ctx=Load()), op=Add(), right=Name(id='d', ctx=Load()))
c + d

BinOp(left=Name(id='g', ctx=Load()), op=Mult(), right=Constant(value=3))
g * 3

Name(id='c', ctx=Load())
c

Name(id='d', ctx=Load())
d

Name(id='g', ctx=Load())
g

Constant(value=3)
3
```

asttokens/asttokens.py
alexmojaki: How about `ASTText` and `ASTTextBase`?

alexmojaki:
@PeterJCLaw if you update this branch with the latest in #93 and consider this PR ok to merge, then I'll probably merge it. No need for a full review here, that can be done within #93.

alexmojaki:
Thanks!

PR #104, opened 2022-12-24
Account for a bug in multiline f-strings

alexmojaki:
I discovered *another* f-string positions bug. It was fixed in https://github.com/python/cpython/issues/94869 and released in 3.10.6 according to https://docs.python.org/3.10/whatsnew/changelog.html#python-3-10-6-final.

Theoretically, you could maybe check if the specific f-string is multiline to continue supporting non-multiline strings between 3.9.7 and 3.10.6.


dsagal (approved):
Thank you!

PR #112, opened 2023-08-06
Add readthedocs configuration file, and fix GHA for Python 2.7

dsagal:
Adds a minimal readthedocs configuration file, to keep https://asttokens.readthedocs.io/en/readthedocs/ active. See https://blog.readthedocs.com/migrate-configuration-v2/.

Also, GitHub Actions has dropped support for Python 2.7 in its built-in actions (see https://github.com/actions/setup-python/issues/672). Add a workaround to keep testing it.


dsagal:
Good catch. Adding PR #115 to fix it.

PR #113, opened 2023-08-07
Update docs building dependencies and configs

dsagal:
- Use pinned and up-to-date dependencies for generating documentation
- Regenerate sphinx config file, and documentation on readthedocs
- Preview available at https://asttokens.readthedocs.io/en/sphinx-conf/


alexmojaki:
The docs preview is missing `ASTText`. Did you build it locally without updating `master`?

dsagal:
The issue is that auto-documentation isn't fully automatic, but only selected classes and members are included. So I have to list ASTText explicitly. Including `:members:` at top level would show that automatically, but in a differently organized layout, so I added ASTText explicitly to keep layout as before. One top-level member not documented is `supports_tokenless`. Not sure it's worth documenting as a top-level function. Maybe could be moved to `util`? Anyway, that seems separate from this task.

alexmojaki (approved):
I don't know what most of this stuff means :shrug: 

Also I still don't see ASTText in the preview, did you rebuild?

alexmojaki:
https://asttokens.readthedocs.io/en/stable/api-index.html still needs updating

dsagal:
I updated a setting, it rebuilt, and looks up-to-date now. It _should_ be automatic, from skimming https://docs.readthedocs.io/en/stable/versions.html. Maybe it was the missing setting. If it doesn't auto-update after next release, I'd have to debug further.

PR #114, opened 2023-08-30
Support ASTText in astroid without tokens

alexmojaki:


dsagal (approved):
Nice! Requesting again some comments in places where it was hard to follow the logic.

PR #117, opened 2023-09-06
Drop python < 3.8

### LINKED ISSUE #111: Drop Python 2.7?
Continuing the discussion in https://github.com/gristlabs/asttokens/pull/110#issuecomment-1615147120

I've just noticed that GHA still has PyPy 2.7, which is actually probably enough to say that the library is still being tested on 2.7 and is thus expected to work. Dropping support (i.e. the PyPI classifier) is only valuable if it saves us from doing work to have that confidence.

On the other hand, I don't want people doing (or thinking that they have to do) extra work in the future for the sake of Python 2, e.g. I feel bad about the effort @palfrey went through with mypy and Python 2.

If we do go forward with this, https://github.com/gristlabs/asttokens/pull/67 would help.

aqeelat:
Closes #111 

Todo:

- [x] Update the tests
- [x] Update the code
- [ ] Add notice to readme and docs
- [ ] (Optional) replace mypy with native typing


asttokens/asttokens.py
```diff
@@ -33,18 +32,17 @@
...
+    # type: (str, str) -> None
     self._filename = filename
 
     # Decode source after parsing to let Python 2 handle coding declarations.
     # (If the encoding was not utf-8 compatible, then even if it parses correctly,
     # we'll fail with a unicode error here.)
-    source_text = six.ensure_text(source_text)
+    # FIXME: This is the only remaining usage for six. Do we still need this check?
+    #  If so, then we should copy the function here and remove dependency on six
+    source_text =  six.ensure_text(source_text)
```
aqeelat, line 45: review this part

tests/tools.py
```diff
@@ -12,8 +12,8 @@
 
 
 def get_fixture_path(*path_parts):
-  python_dir = 'python%s' % sys.version_info[0]
-  return os.path.join(os.path.dirname(__file__), "testdata", python_dir, *path_parts)
+  # TODO: since we don't need python 2 anymore, should we move testdata/python3/astroid to testdata/astroid?
+  return os.path.join(os.path.dirname(__file__), "testdata", "python3", *path_parts)
```
aqeelat, line 16: review this part

aqeelat:
@alexmojaki @dsagal I don't have permissions to add you as reviewers but please look at this pr.

asttokens/asttokens.py
alexmojaki: Yes, I'd say we should still decode bytes to preserve compatibility. So the type hint needs to be amended. Also the "Decode source after parsing to let Python 2..." doesn't even make sense and needs updating.

setup.cfg
```diff
@@ -34,9 +28,7 @@ classifiers =
 
 [options]
```
alexmojaki, line 29: We also need https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

tests/test_asttokens.py
```diff
@@ -146,6 +145,7 @@ def test_unicode_offsets(self):
 
   def test_coding_declaration(self):
     """ASTTokens should be able to parse a string with a coding declaration."""
+    # TODO: figure out if the comment below is relevant
```
alexmojaki, line 148: The Python 2 part can be removed, but let's test that it works with both `bytes` and `str`.

tests/test_mark_tokens.py
```diff
@@ -271,11 +274,12 @@ def test_adjacent_strings(self):
 
 
   def test_print_function(self):
+    # TODO: is this test still needed?
     # This testcase imports print as function (using from __future__). Check that we can parse it.
     # verify_all_nodes doesn't work on Python 2 because the print() call parsed in isolation
     # is viewed as a Print node since it doesn't see the future import
```
alexmojaki, line 280:
```suggestion
    # This testcase imports print as function (using from __future__). Check that we can parse it.
```

tests/tools.py
alexmojaki: it wouldn't hurt, but it's not required

alexmojaki:
Thanks for your hard work, sorry it took so long to respond

tests/test_mark_tokens.py
aqeelat: yeah but why do we need to import print from future?

alexmojaki: This probably isn't testing anything interesting any more but we might as well keep it.

setup.cfg
```diff
@@ -34,9 +28,7 @@ classifiers =
...
 packages = asttokens
-install_requires = 
-    six >= 1.12.0
-    typing; python_version < "3.5"
+install_requires = six >= 1.12.0
```
neutrinoceros, line 31: Doesn't this PR actually make `six` irrelevant and unused ?
```suggestion
```

a-detiste: six usage remains in places

neutrinoceros: I see. I expect applying https://github.com/gristlabs/asttokens/pull/117#issuecomment-1999541944 would solve that, but it could also be done as a follow up patch.

kloczek: Please have look on https://github.com/gristlabs/asttokens/pull/117/commits/de0a2e7307998fea712353fa0865ad0a1b4fec0e where is patch which removes use of six from all places.

PR #120, opened 2023-10-24
Bump urllib3 for security

PeterJCLaw:


PeterJCLaw:
Hrm, I didn't spot that https://github.com/gristlabs/asttokens/pull/119 was already here. We should probably use that.

PR #125, opened 2023-10-24
Clarify the version of astroid which is supported

PeterJCLaw:
This doesn't preclude supporting other versions at some point, however it will make it clearer which versions are supported as well as making CI pass.


PeterJCLaw:
Not sure I understand why coveralls thinks there's a reduction in coverage here!

alexmojaki:
Thanks!

PR #128, opened 2023-10-24
Use pytest-xdist to speed up CI

PeterJCLaw:
Contributes towards #129.
This reduces the CI time from ~30 minutes to ~20 minutes.


alexmojaki:
Because CI has `ASTTOKENS_SLOW_TESTS=1`.

alexmojaki:
Thanks, it's exciting to see this get faster!

PR #132, opened 2023-10-26
Run slow tests without coverage

### LINKED ISSUE #129: CI is absurdly slow
CI seems to take ~30 minutes to run `pytest`, which locally takes about 10 seconds. Do we have any idea what's causing this?

PeterJCLaw:
Per suggestion from @alexmojaki at https://github.com/gristlabs/asttokens/issues/129#issuecomment-1779132057

These slower tests are integration/kitchen-sink tests, so they shouldn't add substantially to the code coverage. Running them without coverage is quite a lot faster (several minutes in CI) which is well worth the small additional complexity.

This commit also drops pytest outputting the junit xml file as there's not an obvious way to combine that from multiple test runs and it appears to be unused.

This reduces the CI time from ~20 minutes to ~10 minutes, however there's a small reduction in _measured_ coverage. Happy to address that here with unit tests if desired.

Builds on #128.
Fixes #129 as CI time is no longer quite as silly.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+branch = True
+include = asttokens/*
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -81,10 +81,16 @@ jobs:
         # pypy < 3.8 very doesn't work
         # 2.7 is tested separately in mypy-py2, as we need to run mypy under Python 3.x
 
-      - name: Test with pytest
+      - name: Fast tests with coverage
         run: |
-          pytest --cov -n auto --junitxml=./rspec.xml
+          pytest --cov -n auto -m "not slow"
           coverage report -m
+
+      - name: Slow tests without coverage
+        # Python 2.7 doesn't have any slow tests
+        if: ${{ !contains(fromJson('["2.7", "pypy-2.7"]'), matrix.python-version) }}
+        run: |
+          pytest -n auto -m slow
 
       - name: Collect coverage results
         env:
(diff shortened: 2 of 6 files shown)


alexmojaki:
Please don't @ me in commit messages, it does weird things with github notifications

PeterJCLaw:
Oh, sorry, hadn't thought about that. Yeah, I suspect it combined badly with force pushing a few times to fix CI.

alexmojaki:
Hmmm, I feel weird about having two mechanisms for dealing with slow tests: ASTTOKENS_SLOW_TESTS and the pytest marker.

PeterJCLaw:
> Hmmm, I feel weird about having two mechanisms for dealing with slow tests: ASTTOKENS_SLOW_TESTS and the pytest marker.

Ah, good point. Do you think that having some of the sys.modules tests in the normal suite adds value? If not, we could just push them entirely into the slow suite and drop `ASTTOKENS_SLOW_TESTS`. On the other hand doing that would mean that locally the default would be to run all the tests, which might not be ideal for rapid iteration. A script which just runs the fast subset might be useful to help there? (Unless there's a way to default `pytest` to not running a particular mark?)

PeterJCLaw:
> Unless there's a way to default `pytest` to not running a particular mark?

There is! We can add `-m 'not slow'` to the `adopts` key in `setup.cfg`. That can still be overridden on the command line (`-m ''` to run everything, `-m slow` to run just the slow ones), so that feels like it's probably ideal here.

Concretely then I'm proposing:
- add `-m 'not slow'` to `adopts` (setting that as the default)
- remove `ASTTOKENS_SLOW_TESTS` completely
- leave CI explicitly setting `-m` as it is now
- maybe document these in the README?

PeterJCLaw:
Have implemented that approach in 52e9b4134ae42818e1dc386af6d9c9905e565016.

alexmojaki:
This looks great, thank you!

PR #134, opened 2023-10-27
Add smoke tests for bytes literals and recursive f-strings

PeterJCLaw:
Coverage of the related codepaths was low within the main unit-test suite, which this aims to help with. These tests are admittedly not great, however should catch any glaring issues.

### DIFF
--- a/tests/test_mark_tokens.py
+++ b/tests/test_mark_tokens.py
@@ -307,6 +307,27 @@ bar = ('x y z'   # comment2
           "%s:%s" % ("AssignName" if self.is_astroid_test else "Name", source.split("=")[0]),
         })
 
+    def test_bytes_smoke(self):
+      const = 'Const' if self.is_astroid_test else (
+          'Constant'
+          if sys.version_info >= (3, 8)
+          else 'Bytes'
+      )
+
+      for source in (
+        'b"123abd"',
+        r"b'\x12\x3a\xbc'",
+      ):
+        expected = {
+          "Module:" + source,
+          const + ":" + source,
+          "Expr:" + source,
+        }
+
+        m = self.create_mark_checker(source)
+        self.assertEqual(m.view_nodes_at(1, 1), expected)
+        m.verify_all_nodes(self)
+
 
   if sys.version_info[0:2] >= (3, 6):
     # f-strings are only supported in Python36. We don't handle them fully, for a couple of
@@ -366,28 +387,6 @@ bar = ('x y z'   # comment2
         )
 
         m.verify_all_nodes(self)
-
-  def test_bytes_smoke(self):
-    const = 'Const' if self.is_astroid_test else (
-        'Constant'
-        if sys.version_info >= (3, 8) or six.PY2
-        else 'Bytes'
-    )
-
-    for source in (
-      'b"123abd"',
-      r"b'\x12\x3a\xbc'",
-    ):
-      expected = {"Module:" + source}
-      if six.PY3 or not self.is_astroid_test:
-        expected |= {
-          const + ":" + source,
-          "Expr:" + source,
-      }
-
-      m = self.create_mark_checker(source)
-      self.assertEqual(m.view_nodes_at(1, 1), expected)
-      m.verify_all_nodes(self)
 
   def test_splat(self):
     # See https://bitbucket.org/plas/thonny/issues/151/debugger-crashes-when-encountering-a-splat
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
Somewhat baffled by the coverage apparently decreasing here!

tests/test_mark_tokens.py
```diff
@@ -338,6 +339,39 @@ def test_adjacent_joined_strings(self):
...
+        if sys.version_info >= (3, 8) or six.PY2
+        else 'Bytes'
+    )
+
+    for source in (
+      'b"123abd"',
+      r"b'\x12\x3a\xbc'",
+    ):
+      expected = {"Module:" + source}
+      if six.PY3 or not self.is_astroid_test:
```
alexmojaki, line 366: It sounds like this is just broken in Python 2 with astroid. We don't have to fix that, but I think it'd be better to just skip this whole test in Python 2, remove the `six.PY2` further above, and collapse this `if` to make things readable. We're supposed to be dropping Python 2.

tests/test_mark_tokens.py
```diff
@@ -338,6 +339,39 @@ def test_adjacent_joined_strings(self):
             "JoinedStr:'x y z'   # comment2\n       'a b c'   # comment3\n       f'u v w'"
         })
 
+  if sys.version_info >= (3, 12):
+    def test_fstrings_3_12_plus(self):
+        m = self.create_mark_checker(
+          'x = (f"Wobble {f"{func(kwarg=f"{boo!r}")}"!r}.",)',
+        )
+
+        self.assertEqual(m.view_nodes_at(1, 6), {
```
alexmojaki, line 348: Can you also check one or both of the inner f-strings?

PeterJCLaw: https://github.com/gristlabs/asttokens/pull/134/commits/1c42c835b0bb525afd2832478c4bf41640027b87 does this to an extent. We don't seem to get the full details of the inner nodes though. I'm not sure if that's expected.

tests/test_mark_tokens.py
PeterJCLaw: I agree the conditionals here aren't great. Note however that dropping support for Python 2 doesn't change the shape of things much -- we still need all the conditions which are present in this code (the conditions just get a little bit simpler). If support for Python < 3.8 goes away too then that'd reduce this to just the astroid differences (in both parts).

alexmojaki: Surely `six.PY3 or not self.is_astroid_test` is always true in Python 3?

tests/test_mark_tokens.py
alexmojaki: Ah, that's sort of expected pre-3.12 but it's also sort of a bug in 3.12 where getting the tokens becomes possible. ASTText can deal with f-strings, but the package is called asttokens...so if you're willing to look into this, search for `include_joined_str`.

tests/test_mark_tokens.py
PeterJCLaw: :facepalm: I somehow wasn't reading the `not`.

tests/test_mark_tokens.py
PeterJCLaw: Yeah, I'd seen that argument to the `iter_children_*` functions. That's the reason for using `node.values` over the `view_nodes_at` approach in the below assertion. Perhaps I'm misunderstanding, but I think the lack of text is coming from the calls to `ASTTokens.get_text_positions` where the implementations don't return useful values for f-strings (due to not having a `first_token`). Checking `ASTText.get_text_positions`'s behaviour might be useful, though I'm assuming that that's covered elsewhere.

I'm not sure what you're suggesting be investigated/potentially changed here?

alexmojaki: I'm suggesting that in 3.12+, `MarkTokens` should start attaching `first_token` to inner f-string nodes.

tests/test_mark_tokens.py
PeterJCLaw: f422904b237b30157d9f04e3ccad8275aa65e9eb

tests/test_mark_tokens.py
PeterJCLaw: Ah, I see. That feels worthwhile, but perhaps a bit out of the scope of this PR? My main goal here was to add some lightweight coverage for things that #132 seemed to show weren't covered by unit tests.

Having a very quick look, just adding `include_joined_str=True` in a few places doesn't seem to pass the tests, so this would probably need a deeper dive really.

alexmojaki: Right, please just add a comment explaining the situation with a TODO, because the assertion is indeed hard to understand.

PeterJCLaw: Is 736d4e3 sufficiently clear?

alexmojaki:
Thanks again for all your contributions!

PR #155, opened 2024-10-06
remove dependency on six

a-detiste:


.github/workflows/build-and-test.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       - name: Mypy testing
         run: |
           # Not an exact mypy version, as we need 0.942 for pypy-3.8 support, but it's not available on 3.5
```
alexmojaki, line 64: "but it's not available on 3.5" also sounds like something that can be fixed now, not necessarily here

setup.cfg
```diff
@@ -30,7 +30,6 @@ classifiers =
 [options]
 packages = asttokens
 install_requires =
-    six >= 1.12.0
     typing; python_version < "3.5"
```
alexmojaki, line 33:
```suggestion
```

alexmojaki (approved):
Great! Is this ready to merge once the tests pass?

.github/workflows/build-and-test.yml
a-detiste: Yes ... but I'm really good at managing the Debian ecosystem and releases; not so much the native Python ecosystem (pip, tox etc...)

a-detiste:
It is ready.

a-detiste:
For what it's worth, a much dumber patch was temporarily used in Debian: https://sources.debian.org/src/python-asttokens/2.4.1-2/debian/patches/remove-six.patch/

The good things is that the CI suite of all reverse dependencies (the one using asttokens) passed fine.

```
-import six
+class six:
+     PY2 = False
+     PY3 = True
+     string_types = (str,)
```

alexmojaki:
Thank you so much!!!

PR #156, opened 2024-11-24
More cleanup of code for Python before 3.8, fix astroid import

alexmojaki:
In preparation for a release I noticed quite a bit of code remaining specifically for old Python versions, as well as an unprotected `astroid` import which would have been a disaster.

View with whitespace changes hidden because big blocks have been unindented: https://github.com/gristlabs/asttokens/pull/156/files?w=1

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,7 +32,6 @@ The API Reference is here: http://asttokens.readthedocs.io/en/latest/api-index.h
 
 Usage
 -----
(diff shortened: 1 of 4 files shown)


dsagal (approved):
All makes sense. Thank you!

alexmojaki:
Thanks! Please can you take care of the release?

PR #157, opened 2025-01-07
Replace ast.Str usages with ast.Constant for compatibility with Python 3.14

befeleme:
Per What's new in Python 3.14:
ast.Str has been deprecated since Python 3.8, and have emitted deprecation warnings since Python 3.12.
https://docs.python.org/dev/whatsnew/3.14.html#id3

Spotted in Fedora Linux while early integrating 3.14 alphas into the distribution. 
I only saw failures in test_util.py, test_asstokens.py; test_tokenless passed successfully in our environment (we've got asttokens 2.4.1). I'm not sure why is that - whether the particular code path is never executed, but I made the change everywhere nevertheless.

### DIFF
--- a/tests/test_asttokens.py
+++ b/tests/test_asttokens.py
@@ -125,7 +125,7 @@ class TestASTTokens(unittest.TestCase):
 
     # Verify that ast parser produces offsets as we expect. This is just to inform the
     # implementation.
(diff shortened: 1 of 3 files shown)


alexmojaki:
Thanks!

PR #163, opened 2025-08-31
Various links improvements

PeterJCLaw:
* Include more links in the package metadata. This makes these slightly more prominent on PyPI.
* Fix tox url
* Make all url literals https: (some were plain http:)


dsagal (approved):
Thank you!

PR #165, opened 2025-11-01
Adapt asttokens for astroid 4.x

### LINKED ISSUE #164: asttokens uses deprecated constant from astroid
The `astroid.Const` is deprecated in favor of `astoid.nodes.Const`. Usage is here:

https://github.com/gristlabs/asttokens/blob/9db9335648659520c7c8453bc39550fa4f95d539/asttokens/util.py#L202C3-L202C28

eendebakpt:
* Replace deprecated usage of `astroid.Const` with `astroid.nodes.Const`
* Add manager argument to `AstroidBuilder` since the default argument was removed on astroid 4.x
* Adapt the tests to treat astroid v4 in the same way as v3

Fixes #164

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,5 +20,5 @@ disallow_untyped_calls=false
 ignore_missing_imports=true
 
 [[tool.mypy.overrides]]
-module = ["astroid", "astroid.node_classes", "astroid.nodes", "astroid.nodes.utils"]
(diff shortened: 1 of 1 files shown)


eendebakpt:
> maybe setup.cfg needs update too?

I created https://github.com/gristlabs/asttokens/pull/166

dsagal:
I pushed a new release. Special thank you to @eendebakpt for the needed maintenance!

PR #172, opened 2026-06-10
Modernize type annotations

AlexWaygood:
This project makes extensive use of legacy type comments for its type hints. These are supported by mypy, but not by more modern type checkers such as ty or pyrefly -- that means that users of the library who want to check their project with ty or pyrefly will get a subpar experience in their IDE, with many variables unnecessarily inferred as `Any`.

This PR modernizes all annotations to use PEP-484/PEP-526 type hints. These should be compatible with Python 3.6+, so easily cover the matrix of Python versions supported by this project.

### DIFF
--- a/asttokens/util.py
+++ b/asttokens/util.py
@@ -38,27 +38,17 @@ from typing import (
 
 if TYPE_CHECKING:  # pragma: no cover
   from .astroid_compat import NodeNG
-
-  # Type class used to expand out the definition of AST to include fields added by this library
-  # It's not actually used for anything other than type checking though!
-  class EnhancedAST(AST):
-    # Additional attributes set by mark_tokens
-    first_token = None  # type: Token
-    last_token = None  # type: Token
-    lineno = 0  # type: int
-    end_lineno = 0 # type : int
-    end_col_offset = 0 # type : int
-
-  AstNode = Union[EnhancedAST, NodeNG]
-
-  TokenInfo = tokenize.TokenInfo
+else:
+  NodeNG = Any
 
 
-def token_repr(tok_type, string):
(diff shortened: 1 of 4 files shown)


asttokens/util.py
```diff
@@ -75,39 +65,49 @@ class Token(collections.namedtuple('Token', 'type string start end line index st
...
-  # type: (Token, int, Optional[str]) -> bool
+# Type class used to expand out the definition of AST to include fields added by this library
+# It's not actually used for anything other than type checking though!
+class EnhancedAST(AST):
+  # Additional attributes set by mark_tokens
+  first_token: Token = None  # type: ignore
+  last_token: Token = None  # type: ignore
+  lineno: int = 0
+  end_lineno: int = 0
+  end_col_offset: int = 0
```
AlexWaygood, line 80: This class was previously under an `if TYPE_CHECKING` block. It could still be under an `if TYPE_CHECKING` block if we made use of `from __future__ import annotations` or extensively quoted various type hints. But that comes at a cost: users wouldn't be able to use [beartype](https://github.com/beartype/beartype) or [typeguard](https://github.com/agronholm/typeguard) at the same time as asttokens, for example, since those libraries require type hints to be resolvable at runtime.

I also doubt the `if TYPE_CHECKING` block was meaningfully speeding anything up here -- one additional class is going to have negligible impact on startup time

AlexWaygood:
The pypy CI failures look unrelated to me

asttokens/util.py
dsagal: This makes sense, thank you.

dsagal (approved):
All looks good. Thank you for this upgrade. 

I agree the pypy failures are unrelated.

